Issue #3, opened 2018-08-17
Add SOFAMosn to Service or Edge Proxy list


rootsongjc:
SOFAMosn is a modular observable smart network which can be used in service mesh deployed as a data plane sidecar. See https://github.com/alipay/sofa-mosn
It is a part of SOFAMesh, see [Introducing SOFAMesh — A solution for large-scale Service Mesh based on Istio open-sourced by Ant Financial](https://medium.com/@jimmysongio/introducing-sofamesh-a-solution-for-large-scale-service-mesh-based-on-istio-by-ant-financial-84ff51fd66f6)

leecalcote:
@rootsongjc thanks for pointing out SOFAMosn! While SOFAMesh leverages Istio's control plane, is based on any other project (outside of reflecting on their design)?

rootsongjc:
@leecalcote Could you add [SOFAMosn](https://github.com/alipay/sofa-mosn) to the website?

leecalcote:
LGTM. Thanks for adding SOFAMosn, @rootsongjc! It seems like SOFAMesh would be appropriate to add as well. What do you think?

PR #4, opened 2018-08-31
Add proxy SOFAMosn

rootsongjc:
Fix issue #3


leecalcote:
LGTM. Thanks for adding SOFAMosn, @rootsongjc! It seems like SOFAMesh would be appropriate to add as well. What do you think?

rootsongjc:
@leecalcote Check this https://github.com/leecalcote/layer5/pull/5

PR #19, opened 2019-01-19
Delete particles.html

leecalcote:
This was just a test page. Now that the particles have been incorporated, this test page can be deleted.


yogi4:
Looks Good

PR #79, opened 2019-02-12
MInor Revision Striped Rows for tables

yogi4:
Minor revision for tables alternating striped rows


yogi4:
Sure lee , I will try that .

PR #80, opened 2019-02-14
Table Color Code by category

yogi4:
Color Coded Tables by category and used striped features for other tables


yogi4:
@leecalcote can you please recommend the right colors for this table

PR #98, opened 2019-03-26
adding the missing info

girishranganathan:


_includes/meshery.html
```diff
@@ -127,6 +129,9 @@ <h5 style="color:#111111;">Run Meshery on Kubernetes:</h5>
 </pre>
     </div>
     <p>If you want to use a different namespace, please change the name of the namespace in the <code>ClusterRoleBinding</code> section appropriately. </p>
+    <p>Access Meshery in your browser at <a href="http://localhost:9081">http://localhost:9081</a>. Configure Meshery with
```
leecalcote: @girishranganathan how to access the meshery ui when deployed in Kubernetes? How are we exposing the service?

girishranganathan: If running on Istio, we do have a gateway and VS configured as part of the Istio.yaml. They will be able to access it through ingress gateway: http://[EXTERNAL IP]:[EXTERNAL PORT]/.
On plain k8s, the meshery service is configured as a service of type load balancer. We could either make the service a NodePort one (OR) include an ingress config. The latter brings in more options which is why I dint go there.

leecalcote: Let's accept for now and revisit K8s later.

PR #102, opened 2019-04-11
updating doc with easy install script steps

girishranganathan:


leecalcote:
@girishranganathan I made a few changes. WDYT?

PR #105, opened 2019-04-21
Posts blog fix

yogi4:


leecalcote:
Thanks, @yogi4. Later, we'll want to expose /blog in the sidebar or footer when posts are actually created. We'll likely look to reduce to either /blog or /posts in the future.

Issue #62, opened 2019-02-01
Include App Mesh in Functional section of landscape


leecalcote:
@subfuzion it'd be great to have App Mesh accurately represented under the [Functional section](https://github.com/layer5io/layer5/blob/master/_data/categories/meshes.yml) of the [landscape](https://layer5.io/landscape/)? Will you consider contributing a PR to ensure its accuracy? Nate Taber assisted with the non-functional already.

leecalcote:
@tabern thank you for your last update to the landscape. It has evolved some. App Mesh details are missing in the [Functional section](https://github.com/layer5io/layer5/blob/master/_data/categories/meshes.yml). Will you consider a PR'ing it?

This landscape will be referenced in three O'Reilly books on the topic of service meshes.

leecalcote:
@nathanpeck will you take a crack at this?

PR #194, opened 2019-10-02
#62 updated Include App Mesh in Functional section of landscape

ashiscs:


_data/categories/meshes.yml
```diff
@@ -10,6 +10,19 @@
   prometheus: "No, Replaced by Harmony Controller"
   tracing: "Yes"
   encryption: "Yes"
+  
+  - name: App Mesh
+  desc: A multi-service mesh management plane to enable adoption, operation and development of service meshes and services running on them
```
leecalcote: @ashiscs thanks for submitting this PR. The description used here is that of the Meshery project. Will you find and use a description for App Mesh instead?

_data/categories/meshes.yml
```diff
@@ -10,6 +10,19 @@
...
+  link: https://github.com/layer5io/meshery-app-mesh
```
leecalcote: We'll want for the link to point to AWS App Mesh, not the Meshery adapter for App Mesh.

leecalcote:
@ashiscs thanks for submitting this PR. The description used here is that of the Meshery project. Will you find and use a description for App Mesh instead?

PR #196, opened 2019-10-02
Added SuperGloo to landscape tool section #110

gunjanmimo:


bettyjunod:
Yes to add to the tools section as this is not a service mesh but an orchestrator for meshes

PR #226, opened 2019-11-01
add Go kit to client library list

basvanbeek:
Adds Go kit to client library list.

Not sure if it makes sense to place Go kit under Service Mesh offerings although it is certainly possible to create a service mesh with Go kit using its resiliency, service discovery, observability and load balancer packages but it is up to the package consumer to wire it as such.


leecalcote:
@basvanbeek thank you!  

You're right... gokit being represented under Client Libraries as a predecessor to Service Meshes is a more accurate, logical categorization than gokit directly considered (categorized) as a service mesh.

Issue #230, opened 2019-11-05
Mesheryctl: Installation using brew


leecalcote:
**Current State:**
The [getting-started](https://layer5.io/meshery/#getting-started) section describes how to install the `meshery` bash script.

**Desired State:**
A new `mesheryctl` binary has been created and is available to install using `brew` on Linux and MacOS. The documentation should describe how to install `mesheryctl` using brew as described in here - https://github.com/layer5io/homebrew-tap/blob/master/README.md.

The method of installing and using `meshery` bash script should continue to be available for now, but hidden by default using ```class="toggle"``` akin to how the Kubernetes instructions are hidden by default.

---
**Contributor Resources**
- [Meshery documentation site](https://meshery.layer5.io/docs/)
- [Instructions for contributing](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)

leecalcote:
@gunjanmimo if you take this issue, you can use [docs/installation](https://github.com/layer5io/meshery/blob/master/docs/pages/installation/index.md) as reference.

PR #240, opened 2019-11-19
small changes - brew instructions

gunjanmimo:


gunjanmimo:
fixes #230

leecalcote:
> @gunjanmimo the brew instructions are to be in addition to the curl instructions. The brew instructions are an alternative way to get started with Meshery. Will you refactor this PR, so that both instructions are included?

@gunjanmimo these two sets of instructions should be separate from one another.

Issue #287, opened 2020-01-21
All pages: Extraneous section on righthand side


leecalcote:
**Description**
On the righthand side of each page on layer5.io, there is an extraneous section of the UI that causes a horizontal scrollbar to appear in the user's browser and allows the users to scroll to the right.

**Expected Behavior**
This unused section of the UI should be removed.

**Screenshots**
<img width="658" alt="Screen Shot 2020-01-21 at 5 03 36 AM" src="https://user-images.githubusercontent.com/7570704/72799776-91676280-3c0b-11ea-8114-d76e963d51c9.png">

leecalcote:
This is likely caused by the rich tooltips used in the social buttons in the footer of the site.

arghac14:
@leecalcote Can I work on this?

leecalcote:
@arghac14 that would be great, yes.

leecalcote:
@arghac14 note that @gunjanmimo has a helpful PR #290 that removes visibility of the horizontal scrollbar. This does a good job of addressing the symptom, but doesn't quite address the root cause, yet.

arghac14:
@leecalcote but how do I run this site locally on Windows platform? Not working using 'make site'

leecalcote:
@arghac14 Good question. You have a couple choices. One is to use Docker and run `make docker` (or execute the commands listed under the `make docker` stanza in the Makefile). You'll need to have Docker installed on your Windows machine. Another option install Jekyll on your Windows system directly. [This tutorial](https://www.youtube.com/watch?v=uxElyNDWoO8) uses Chocolatey (a package manager for Windows) to into install Jekyll.

leecalcote:
@arghac14 you are welcome to jump in the community Slack (http://slack.layer5.io) and/or either the weekly community calls where we troubleshoot and work through issues like this.

leecalcote:
Thank you, @NupurThakur27 !

PR #290, opened 2020-01-21
#287 the scrollbar is removed from layer5.io

gunjanmimo:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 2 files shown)


leecalcote:
Thank you @gunjanmimo. This change is helpful in that removes visibility of the horizontal scrollbar. This does a good job of addressing the symptom, but doesn't quite address the root cause, yet.

If/when the root cause is addressed, it makes sense to come back to remove this code (to not hide the horizontal scrollbar, which is useful on certain pages, like the landscape).

Issue #300, opened 2020-01-23
Add Microsoft and SMI to the list of partners


leecalcote:
**Current Behavior**
Layer5 is partnered with Microsoft and [SMI](https://smi-spec.io)

**Desired Behavior**
This partnership should be represented in [/partners](https://layer5.io/partners)

ashiscs:
Ready to be closed!!

PR #310, opened 2020-01-26
#300 microsoft and smi are added as technology partners

gunjanmimo:


ashiscs:
fixes #300

Issue #288, opened 2020-01-21
Multiple Language Support


leecalcote:
**Current Behavior**
This site currently supports English as its only language.

**Desired Behavior**
Ideally, this site's resources are available in multiple languages (e.g. Spanish, Tamil, Chinese). 

Consider use of https://github.com/kurtsson/jekyll-multiple-languages-plugin

gunjanmimo:
@leecalcote sir, I have doubt that if we are using Jekyll Multiple Languages Plugin, then every time when layer5 will get any post or blog or etc, then we will need to make translation of every post in other languages also. So I made something like this which can translate the whole page in one click.
![Screenshot from 2020-01-27 21-59-14](https://user-images.githubusercontent.com/46948185/73193480-a171cb80-4150-11ea-8115-f5dfee5d2926.png)
so for our convenience, should I do this? 
![screencapture-127-0-0-1-4000-blog-2020-01-27-22_03_10](https://user-images.githubusercontent.com/46948185/73193612-dd0c9580-4150-11ea-90b4-1e09455a3d26.png)

PR #313, opened 2020-01-27
#288 multi language support for layer5.io

gunjanmimo:
@leecalcote sir, I have made these changes, if we need something further then I have a plan that is to detect the default browser language of the user and it will automatically translate it.  and I have added this language change section if a user likes to read layer5.io contents in their language.


ashiscs:
fixes #288

leecalcote:
@gunjanmimo interesting. Yes, I think this could work instead. Can you:

1. ensure all icons retain their aspect ratio, so as to avoid a squishing affect?
1. restyling the svg or png of each flag to be monochromatic (akin to the buttons under the **Contact** section)?

gunjanmimo:
@leecalcote sir, is it correct now?

leecalcote:
Looking good. Let's see how the flags look with only width or height statically specified.

Issue #321, opened 2020-01-31
Add multiple languages to the layer footer.


Aisuko:
**How can we help?**
A brief description of what question(s) you have and how we can help clarify.

As @leecalcote  said in the channel, we could add Chinese into [layer.io](https://layer5.io) footer.

Aisuko:
@gunjanmimo So, Would you like to tell me where you got the national flags? I will get one same size as other flags

jijillery:
The SVGs for these flags were found on Wikipedia ([flag of China](https://en.wikipedia.org/wiki/Flag_of_China#/media/File:Flag_of_the_People's_Republic_of_China.svg)). Not only are these nice SVGs, but they carry a public domain copyright for our use.

Aisuko:
As I discussed with @gunjanmimo. I can confirm add Chinese and Japan flags to the site.

PR #324, opened 2020-02-01
Add multiple language

Aisuko:
Signed-off-by: Aisuko <urakiny@gmail.com>

@gunjanmimo As we discussed. Add Chinese and Japanese flags.

- fixes: #321 

Language support code reference. https://cloud.google.com/translate/docs/languages


gunjanmimo:
it is a great work @Aisuko, thank you

PR #349, opened 2020-03-08
added language flags

### LINKED ISSUE #346: Add language flags to `footer-plain.html`
**Current Behavior**
Layer5.io has two footers. One of these footers is without the translation capability, which means that if a user has set the site to a certain language and then navigates to a page that is using the footer that doesn't have the language flags (doesn't support translation), that page's content is not translated to the language of the user's choice.

**Desired Behavior**
Add translation capability to the other footer (`footer-plain.html`).

ashiscs:
fixes #346 


leecalcote:
Thank you, @nupur-thakur!

PR #360, opened 2020-03-16
updated landscape section

ashiscs:


ashiscs:
@leecalcote sir ready for review..

leecalcote:
@ashiscs thanks for taking a crack at this update. @rootsongjc food for thought - instead of removing SOFAmesh altogether, we could mark it as archived similar to what was done for the Rotor service mesh. In this way, the landscape is more complete with a record of the history of the ecosystem showing the organic nature of service meshes coming and going. What do you think?

rootsongjc:
@leecalcote Sounds good to me. Leave trace for the history of Service Mesh.

PR #365, opened 2020-03-18
fixes the bug present related to the link

### LINKED ISSUE #364: [News] What the Numbers Say about How Service Meshes Are Used Today
**Description**

The last news item, there is a small bug. If you click on the image in the card, the link is https://layer5.io/news/what-the-numbers-say-about-how-service-meshes-are-used-today, but needs to point to The New Stack (like the small, blue circle button does).

ashiscs:
fixes #364 


ashiscs:
@leecalcote sir, is this okay?

leecalcote:
@ashiscs hmm. This behavior isn't quite what we're looking for. It seems that we're going to need to update the logic in https://github.com/layer5io/layer5/blob/master/_includes/collections/news.html to ensure that when an `eurl` is present that the hyperlink for the image and for the button both direct the user to an external site.

PR #368, opened 2020-03-18
added languages.html

ashiscs:


leecalcote:
I should have noted that if we haven't already, we can use this same `{% include .... %} in the `_includes/footer.html` file as well.

PR #379, opened 2020-04-03
Git-ignore should not be present in git-ignore

kushthedude:
Gitignore may tend to change hence it should be present in git-ignore


kushthedude:
@leecalcote Netlify is in action here too. Every commit to master will trigger a build and publish the site to https://layer5io.netlify.com

sachincool:
I love it ❤️@kushthedude Great work!

Issue #385, opened 2020-04-23
Prepare for participation in Season Of Docs


kushthedude:
After discussing the potential opportunity for Layer5 to participate in GSoD. 

I am glad to announce, Layer5 as an organisation is applying for **[Google Season of Docs 2020](https://developers.google.com/season-of-docs)**.

This will be a tracking issue for the participation of the community in the GSoD -:
- [ ] Prepare the potential idea list for GSoD.
- [ ] Create a GSoD page on layer5.io which will contain information about the project.
- [ ] Apply for the GSoD as a mentoring organisation.

----

Additional Resources -:

- [Timeline](https://developers.google.com/season-of-docs/docs/timeline)
- [Application Guidelines](https://developers.google.com/season-of-docs/docs/organization-application-hints)
- [Information about Ideas & Publishing Publicly](https://developers.google.com/season-of-docs/docs/project-ideas)

leecalcote:
@kushthedude as we go to close out this issue, will you ensure that we have captured our program proposal in a google doc?

kushthedude:
> @kushthedude as we go to close out this issue, will you ensure that we have captured our program proposal in a google doc?

We already have captured our proposal in google doc, as far as I remember.

leecalcote:
Found it! Thank you, @kushthedude. Let's earmark the calendar for participation next year.

PR #388, opened 2020-04-24
feat: Idea List for GSoD-2020

kushthedude:
Refers #385


kushthedude:
Migrate to `/gsod/2020` for GSoD page. I have not created a hyperlink as of now since we are not sure of our acceptance.

PR #395, opened 2020-04-29
Minor punctuation correction

kushthedude:


kushthedude:
Since it's a minor neat, I made this PR from Github UI. Therefore, the DCO issue is there if you say I will make the change from local.

PR #402, opened 2020-05-01
Prettify all HTML files

aayushdutt:
Currently, the repository is having a huge number of unformatted HTML files. This makes it almost impossible to find missing tag elements or find extra HTML tags. 
In this PR I have just run a prettier prettify on all the HTML files of the repository. Since it only changes whitespace, it should not break anything. I have manually tested some pages and they were working just as on layer5.io
Signed-off-by: Aayush Dutt <aayushd100@gmail.com>


aayushdutt:
Also, while prettifying the html files, I saw that there is a huge number of HTML validation errors. I aim to fix those next.

aayushdutt:
> Thank you, @aayushdutt.
> 
> Is this prettifier something we might consider including in the build?

@leecalcote that will be perfect! This will ensure the files remain properly formatted and valid. I'll go ahead and create an issue.

Issue #407, opened 2020-05-02
Setup prettier


aayushdutt:
**Current Behavior**
The repository contains a lot of invalid and unformatted files. This makes it hard to maintain.

**Desired Behavior**
It will be great to set up prettier (and maybe husky) to ensure the files are formatted and valid.

PR #410, opened 2020-05-08
#407 prettier and eslint setup with husky precommmit hook

gunjanmimo:


gunjanmimo:
> Considering the absence of JavaScript Files, also since it is a Jekyll based site we only need to lint our markdown files. ESLint & yarn engine makes no sense for a Jekyll blog.
> We should use https://github.com/markdownlint/markdownlint , written in ruby specifically for Jekyll based websites.

@kushthedude I have just tried Markdown lint tool, and it is suitable for layer5 repo. But I have doubts. Markdown lint tool is only showing the errors, how to resolve those errors automatically? Is there any tool like prettier? and do we need recommit hook?

kushthedude:
Pre-commit hook will be difficult without presence of node engine. I dont
know if there is any gem with support of pre-commit hook. Coming to
prettier, these tools are compatible only with node engines I dont think
there is any alternative gem.

leecalcote:
Does anyone have a suggestion as to what action to take on this PR?

Nikhil-Ladha:
As oer the last discussion, I think we need a markdown lint rather than ESlint ehich is not provided by this PR. So, we may close this one now.
Moreover, as we are moving to the layer5 site there we can use ESlint. So, we can add the linting feature after that site is ready.

PR #440, opened 2020-06-09
minor tweak in the footer and the careers post

ashiscs:

### DIFF
--- a/assets/css/materialize.css
+++ b/assets/css/materialize.css
@@ -2262,7 +2262,11 @@ li.youtube {	list-style-image: url('/assets/images/youtube.svg'); }
 
 .black-text {
   color: #000000 !important;
-  font-size: medium;
+}
+
+.mlh {
+  color: #000000 !important;
+  font-size: 14px;
 }
 
 .white {
--- a/assets/css/materialize.css
+++ b/assets/css/materialize.css
@@ -2262,6 +2262,7 @@ li.youtube {	list-style-image: url('/assets/images/youtube.svg'); }
 
 .black-text {
   color: #000000 !important;
+  font-size: medium;
 }
 
 .white {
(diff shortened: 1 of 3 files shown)


collections/_careers/product-engineering-internship.html
```diff
@@ -33,9 +33,9 @@ <h5 class="black-text">What You Will Learn</h5>
 Your work will include designing, coding, testing, and deploying cloud native solutions. 
 
 We have multiple internships for the Spring and Summer. 
-This is an extremely dynamic position and a great opportunity for an entry level marketing person to build their skills and their career.
+This is an extremely dynamic position and a great opportunity for an entry level tech person to build their skills and their career.
 The Software Engineer Intern will work closely with the open source community and broad set of Layer5 academic and technology partners.
-The ideal candidate is enjoys learning at a high rate and is driven to succeed with little supervision. 
+The ideal candidate is who enjoys learning at a high rate and is driven to succeed with little supervision.
```
leecalcote, line 38: Thank you!

assets/css/materialize.css
```diff
@@ -2262,6 +2262,7 @@ li.youtube {	list-style-image: url('/assets/images/youtube.svg'); }
 
 .black-text {
   color: #000000 !important;
+  font-size: medium;
```
leecalcote, line 2265: I’m concerned about this change. This class Is used in a number of places. I haven’t looked yet to see how those different places are affected.

Nikhil-Ladha: This shouldn't be done IMO. As this class is used in the nav section too, where on small screen there can be issues with the breaking of UI. And in other places too,. If you have checked all parys then, it's okay.

ashiscs: Okay will verify it again and update if required !!

leecalcote:
Thank you, @ashiscs!

PR #443, opened 2020-06-09
Merge with upstream master, finalize landscape spectrums styles update

aayushdutt:


leecalcote:
@aayushdutt on an unrelated note, you're right about the hover action on the material ui buttons. It would be better if the lifted like you pointed out before.

aayushdutt:
@leecalcote Yeah sure, pushing fixes for these issues
1. Hover animation on buttons
2. Popup overlap issue

aayushdutt:
@leecalcote I have pushed the fixes for the above issue.
Apart from the above issues, I have also changed the color of the buttons throughout the website.
Since this is a big visual change, let me know if I should revert back to the older color

leecalcote:
I see blue as the color of the buttons still. Is it the color of the cards on the landscape that you're highlighting as changed? `master` is still a few commits ahead of this branch. I'll pull in those changes, assuming that doesn't mess you up...

leecalcote:
@aayushdutt ok, I see now - we went from blue to green. Got it. I had tried this combo in the past and felt that it was too much of the same; too little variation. Imagine that we were to finish this transition with a change of hyperlink color, too, from blue (see links on the landscape) to green as well. My sense it that it would be overkill (too much green) and that what we ultimately need is another color (perhaps, one that we don't have today). Hmm.

aayushdutt:
@leecalcote
Yup I agree that there will be too much green. No issue with the blue we are using, just that it's a bit too muted. Let's maybe revert this color change.

aayushdutt:
Also, for the too much green part. There are actually many great websites which use monochromatic color scheme (black, white, and different shades of primary color).

leecalcote:
@aayushdutt to your point, I wonder if the colors found in the footer of the Layer5 document template might have what we're looking for.

PR #510, opened 2020-07-21
fix: Remove double rendering of projects from performance

kushthedude:
https://layer5.io/performance see the double rendering


kushthedude:
@leecalcote I guess this could get in on priority.

PR #532, opened 2020-07-26
Basic design of Footer Component

### LINKED ISSUE #503: [NG] Migrate content of existing Footer to new Footer component
**Current Behavior**
The content of the existing site [footer.html](https://github.com/layer5io/layer5/blob/master/_includes/footer.html) needs to be migrated to the layer5-ng site.

**Desired Behavior**
Layer5-ng site's footer should have almost all the same content.

**Implementation**
1. Update colors to match [Layer5's brand](https://layer5.io/brand)
1. Make twitter feed a separate component; embed twitter feed component into new footer.
1. Make language flags a separate component; embed language flags component into new footer.

Jashpatel1:
**Description**

This PR fixes #503

**Notes for Reviewers**


Jashpatel1:
@leecalcote and @Tanuj22 This is the basic structure for Footer, if this structure is approved, I will proceed to detailing part of the footer.

src/sections/Footer/language.style.js
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components';
+
+const Languagestyle = styled.section`
```
Tanuj22, line 3: `LanguageWrapper` would be a better name to maintain consistency

Jashpatel1:
@leecalcote Please confirm regarding the social and projects section as it was intended to reduce redundancy as projects section is already present in header and social would be included in the link icons. Rest a few changes have been made as stated by Tanuj and Nikhil.

Nikhil-Ladha:
@Jashpatel1 few more changes needed.
All the flags should be if same width and height. There is some issue with images of flag, please confirm that. For example, the flag of India is in black color.
Languages and Contact should be changed to a header format or in whatever form the other optios are in the footer.
We can remove the footer options of the theme, which are present by default and also add the copyright for Layer5.
And the twitter iframe should be at the bottom, for mobile view. If that's possible.

leecalcote:
Good progress here, @Jashpatel1. 👍 

Here's a screenshot of what @Nikhil-Ladha noted with one of the flags. 
Screen Shot 2020-07-28 at 4 14 23 PM
 
We might want to have the Languages section title in bold to follow suit with the others.

Jashpatel1:
@leecalcote and @Nikhil-Ladha I have made significant changes in the layout as well as fixed all of the issues mentioned by you except the twitter iframe being at bottom in mobile view. Please have a look.

Tanuj22:
@Jashpatel1 I am seeing an overflow of the flags
Screenshot from 2020-07-30 18-22-28

Nikhil-Ladha:
> @Jashpatel1 I am seeing an overflow of the flags
> Screenshot from 2020-07-30 18-22-28

Just what I was gonna mention.
Though another part, due to the margins specified in the language and contacts section the screen is overflowing on the right-hand side, adding white space. `Fix: Try removing the margin and add padding instead, as there you are using width:100% which is the reason for the overflow`

Nikhil-Ladha:
Also, I think we agreed on shifting the tweet in the bottom and having the contact section beside the logo? Similar, to the hashicorp site?
If we do that, I would suggest using the smaller logo of the site as there in the header as this one looks a bit odd and big.

Jashpatel1:
> Also, I think we agreed on shifting the tweet in the bottom and having the contact section beside the logo? Similar, to the hashicorp site?
> If we do that, I would suggest using the smaller logo of the site as there in the header as this one looks a bit odd and big.

In case we want to do that, we will have to change the layout then, and I think this is the most compact layout set, in case we bring the tweet section down we will have empty spaces as we have in current layer5 site. So, please let me know as i dont recall if it was decided in the meeting. If it has to be done, please let me know.

Nikhil-Ladha:
What do other's suggest?
// @leecalcote @Tanuj22 @dhruv0000 @kushthedude ?

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+              </Row>
+          </Col>
+        </Row>
+          </Col>
+        </Row>
+        <div className="footer-bottom">
+          <ul className="footer-menu">
+            <li>
+              <Link className="anchor" to="#">
+                Privacy
```
kushthedude, line 201: Can they be populated dynamically using sitemap?

Tanuj22: Do you mean the sitemap plugin for gatsby?

dhruv0000:
The current websites footer has a section for the `projects`, which is missing here. What if we use the space occupied by the tweets to highlight them (using logos), like HarshiCorp did with its footer?

src/sections/Footer/index.js
kushthedude: Yes I guess, data should be populated from the sitemap rather than hardcoding. But if that is not feasible right now we can move ahead with the PR as a first iteration.

Nikhil-Ladha:
Please make set the same width for all the flags.
Also, if we could add the hover effect in the contact section as we have in the current site it would be better.
And, the languages and contact section is not properly aligned for mobile view.

Jashpatel1:
> Please make set the same width for all the flags.
> Also, if we could add the hover effect in the contact section as we have in the current site it would be better.
> And, the languages and contact section is not properly aligned for mobile view.

Could you let me know in what sense the alignment is not proper, as I am unable to recognise it.

Nikhil-Ladha:
Screenshot_20200801-222038
You can see the languages and contact section is shifted a bit towards left.

Nikhil-Ladha:
I think you can actually shift the above columns as well to a bit left,rather than shifting the languages and contact to a bit right.

leecalcote:
@Jashpatel1 I wonder if we should be using Gatsby's official plugin for the Twitter feed - https://www.gatsbyjs.org/packages/gatsby-plugin-twitter. I ask, because it would naturally seem to be better integrated and better sustained. Are there any other considerations we should make for for using https://www.gatsbyjs.org/packages/gatsby-plugin-twitter/ vs https://www.npmjs.com/package/react-twitter-embed?

src/sections/Footer/footer.style.js
```diff
@@ -1,199 +1,236 @@
...
-          margin-top: 30px;
-        }
-      }
-    }
-  }
-`;
-
-export default FotterWrapper;
+import styled from 'styled-components';
+import FooterBG from '../../assets/images/app/footer/footer-bg.png';
```
leecalcote, line 2: Can we reduce the size of this background image to 1 pixel and have it repeated horizontally and vertically instead, so that we can achieve faster load time and higher page rank? Even better, can we just use CSS to color the background?

More so than either of these two comments, can we get back to something similar to the original theme? If not the original theme, then something with more activity (more curves and gradients)?

Screen Shot 2020-08-02 at 10 03 02 PM

Screen Shot 2020-08-02 at 10 00 53 PM

src/assets/images/icons/github.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 400 400">
+    <path
+    fill="currentColor"
```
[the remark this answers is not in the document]
leecalcote: Relocated.

leecalcote:
@Jashpatel1 thank you for taking on the footer. Can we try to get a dynamic color fill effect like that of the current layer5.io site or that of the appion theme?

src/assets/images/flags/it.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
leecalcote, line 1: Can we change the flags back to grayscale? Full colors are too bold and too eye-catching.

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+                <Link className="social-icons" to="#">
+                  <img height="40 px" src={slack_icon} />
+                </Link>
+                <Link className="social-icons" to="#">
+                  <img height="40 px" src={twitter_icon} />
+                </Link>
+                <Link className="social-icons" to="#">
+                  <img height="40 px" src={github_icon} />
+                </Link>
+                <Link className="social-icons" to="#">
```
leecalcote, line 186: Given that each of these links are to external locations, we can go ahead an include the actual links now for each of the buttons.

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+        <div className="footer-bottom">
+          <ul className="footer-menu">
+            <li>
+              <Link className="anchor" to="#">
+                Privacy
+              </Link>
+            </li>
+            <li>
+              <Link className="anchor" to="#">
+                Terms
```
leecalcote, line 206: See related issue #https://github.com/layer5io/layer5/issues/557

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
+                    Internships
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
+                    GSOC 2020
```
leecalcote, line 118: Let's lowercase the "o" in GSoC

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+                    Service Mesh Tools
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
+                    Service Mesh Books
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
```
leecalcote, line 80: This needs to be collapsed with line 85 to be "Service Mesh Workshops". The link should be to `/workshops`

src/sections/Footer/index.js
```diff
@@ -1,170 +1,219 @@
...
+                    Service Mesh Comparison
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
+                    Service Mesh Tools
+                  </Link>
+                </li>
+                <li>
+                  <Link className="anchor" to="#">
```
leecalcote, line 75: Link to `/books`

Jashpatel1: This can be done once the link gets active, correct?

src/sections/Footer/index.js
Jashpatel1: The link can be done once we have a page there.

src/sections/Footer/index.js
leecalcote: @Jashpatel1, yep, we can wait on those, if you like. It's ok if they are broken for the time being. My thinking here is that having the links present there leaves one less item undone work item. But, either way will work.

src/sections/Footer/index.js
leecalcote: Good suggestion about the sitemap. It'll be too dynamic for the footer, though, considering how sprawling the site will be. The footer will need manual curation.

src/sections/Footer/index.js
leecalcote: Thanks for collapsing these into one. Looking good. 

Here's a pesky item that you might give some consideration to:

Screen Shot 2020-08-02 at 11 21 42 PM

src/sections/Footer/index.js
Jashpatel1: Updated the links

src/sections/Footer/index.js
Jashpatel1: Updated.

src/sections/Footer/index.js
Jashpatel1: Updated all.

Nikhil-Ladha:
I think now almost all the necessary comments have been addressed and we are good to go with this PR. Any other enhancements can be done through future updates. We are good to go here:+1:

Jashpatel1:
> I think now almost all the necessary comments have been addressed and we are good to go with this PR. Any other enhancements can be done through future updates. We are good to go here:+1: 

Yes, I was going to address this.

Nikhil-Ladha:
Basic design has been changed to intermediate in this PR 😆.

Issue #519, opened 2020-07-24
[NG] Migrate blogs from layer5 to layer5-ng


Jashpatel1:
**Current Behavior**
The current layer5-ng has few sample blogs for reference.

**Desired Behavior**
Migrate blogs from [layer5](https://layer5.io/blog) to the blogs section under [layer5-ng](https://github.com/layer5io/layer5/tree/layer5-ng) using the filesystem-plugin.

---

Jashpatel1:
@anirudhjain75 It would be great help if you would tackle this issue.

chandrashritii:
@anirudhjain75 Thank you for taking an interest in the project 🎈 If you do decide to take this up, take a look at the #496 PR, to get a better idea of what needs to be done.

Jashpatel1:
All the blocks have been added.

PR #545, opened 2020-07-30
Migrate 2 blogs to layer5-ng

Tanuj22:
**Description**
Related to #519 
- Changed folder structure
- Removed inline styles from blogs 
- migrate two blogs

**Notes for Reviewers**


Tanuj22:
@Nikhil-Ladha shouldn't the title of the blog be shown in full? How about fixing the size of the card to a particular value?

Nikhil-Ladha:
> @Nikhil-Ladha shouldn't the title of the blog be shown in full? How about fixing the size of the card to a particular value?

We can, but may be in near future we have a blog with a bigger title, then? We will again set a new height?
I also think it as if a person who is new to the site comes to check out the site, will be curious to know what the topic is and may even just going through all the blog previews, even click on one and see it. Just a nasty thought though 😆.

Tanuj22:
Okay! Changing it right away :satisfied:

PR #547, opened 2020-07-31
Updated issue templates

### LINKED ISSUE #524: Add new issue template for Layer5-ng contributions
**Current Behavior**
The current issue templates assume work will be done on the `master` branch.

**Desired Behavior**
This repo needs an issue template that provides `layer5-ng` specific instructions.

e.g. 

```
---
**Resources**
This issue pertains to the next-generation of the layer5.io website. The site uses Gatsby and Strapi.
- See [`layer5-ng` contributing instructions](link here)

```

Ruth-ikegah:
Signed-off-by: Ruth-ikegah <ruthikegah@gmail.com>

**Description**

This PR fixes #524

**Notes for Reviewers**


Nikhil-Ladha:
I think the issue was to add a new template specifically for layer5-ng site and not to update all the existing templates. Correct me if I am missing something here @leecalcote.

Ruth-ikegah:
> I think the issue was to add a new template specifically for layer5-ng site and not to update all the existing templates. Correct me if I am missing something here @leecalcote.

If this is the case, @leecalcote will have to configure a new template from his end because I don't think I have access to do it. Then I populate the template with contents

Ruth-ikegah:
> It would be good if we add the layer5-ng tag by default in this templates.

a good idea

Nikhil-Ladha:
> > I think the issue was to add a new template specifically for layer5-ng site and not to update all the existing templates. Correct me if I am missing something here @leecalcote.
> 
> If this is the case, @leecalcote will have to configure a new template from his end because I don't think I have access to do it. Then I populate the template with contents

As far as I know, for that you only need to add a new file with a almost similar name and then add the details inside it for the template. Github recognises it on its own.

Ruth-ikegah:
> > > I think the issue was to add a new template specifically for layer5-ng site and not to update all the existing templates. Correct me if I am missing something here @leecalcote.
> > 
> > 
> > If this is the case, @leecalcote will have to configure a new template from his end because I don't think I have access to do it. Then I populate the template with contents
> 
> As far as I know, for that you only need to add a new file with a almost similar name and then add the details inside it for the template. Github recognises it on its own.

okay, thanks for this correction.

Jashpatel1:
@leecalcote  Wasn't it to be done in the layer5-ng branch?

leecalcote:
@Jashpatel1 argh. You're right, yes. I was trying to move too quickly here. I'll revert.

PR #553, opened 2020-08-01
feat: migrate three blogs into layer5-ng

Tanuj22:
**Description**

Related to #519 

**Notes for Reviewers**


Tanuj22:
@Nikhil-Ladha It is being done by Gatsby automatically. If I let it fill the let only `max-width : 500` and let the height be auto. It is causing the same problem of cards being of different sizes. Should  I manually change the dimensions of the image so that Gatsby doesn't crop it.

Nikhil-Ladha:
We can use a wrapper around the image part of the blogs and then apply css to it, to fix that.

Tanuj22:
I am having a bit of trouble with this. Can you check it once @Nikhil-Ladha

Nikhil-Ladha:
There is a class in the  tag inside post-thumb-block, named 'gatsby-image-wrapper'. It has few inline stylings, is that done by us or gatsby itself? If done by us then, you can add an attribute named `height:100%` to it. And, change the object-fit to cover. It will resolve as far as I can see from the developers console.

Tanuj22:
> There is a class in the  tag inside post-thumb-block, named 'gatsby-image-wrapper'. It has few inline stylings, is that done by us or gatsby itself? If done by us then, you can add an attribute named height:100% to it. And, change the object-fit to cover. It will resolve as far as I can see from the developers console.

'gatsby-image-wrapper' is being added by gatsby. I can't figure out any way to add any style to it. Also If I can `height: 100%` to it and make `object-fit: cover` isn't it leading to cropping of image

Jashpatel1:
> > There is a class in the  tag inside post-thumb-block, named 'gatsby-image-wrapper'. It has few inline stylings, is that done by us or gatsby itself? If done by us then, you can add an attribute named height:100% to it. And, change the object-fit to cover. It will resolve as far as I can see from the developers console.
> 
> 'gatsby-image-wrapper' is being added by gatsby. I can't figure out any way to add any style to it. Also If I can `height: 100%` to it and make `object-fit: cover` isn't it leading to cropping of image

Same problem.

Nikhil-Ladha:
> > There is a class in the  tag inside post-thumb-block, named 'gatsby-image-wrapper'. It has few inline stylings, is that done by us or gatsby itself? If done by us then, you can add an attribute named height:100% to it. And, change the object-fit to cover. It will resolve as far as I can see from the developers console.
> 
> 'gatsby-image-wrapper' is being added by gatsby. I can't figure out any way to add any style to it. Also If I can `height: 100%` to it and make `object-fit: cover` isn't it leading to cropping of image
> 
> Edit: Was able to add `height: 100%` to 'gatsby-image-wrapper' but making `object-fit: cover` is leading to cropping of image.

Strange, I tried it and it was working fine for me. Will check again.

Nikhil-Ladha:
blogs
See for me, it works fine.
But, I had a doubt, that if we add the CSS to the `gatsby-image-wrapper` class explicitly, and gatsby uses it somewhere else then it will be a issue later on.

Tanuj22:
object-fit:cover would lead to a problem in image like these :sweat_smile:
Screenshot from 2020-08-02 20-12-49

I am adding css to `gatsby-image-wrapper` only in this case so it won't lead to a problem in other cases.

PR #556, opened 2020-08-02
feat: add react-prism-renderer and move three blogs 

Tanuj22:
**Description**
- Add react-prism-renderer to allow code highlighting and copy code to clipboard
- Migrate three blogs

Related to #519 

**Notes for Reviewers**


src/posts/blogs/Blog.style.js
```diff
@@ -43,7 +32,7 @@ export const BlogWrapper = styled.div`
       border: 2px solid gray;
     }
     .text-centre {
```
leecalcote, line 34: I wonder if we might move away from this British spelling of "center" at some point.

Nikhil-Ladha:
Hi @Tanuj22 and @leecalcote there is some issues with this PR which needs to be addressed in another PR. The issue is with the mobile view of the blog, as the `Copy` button is not fixed in the clipboard and it moves with the text.
Screenshot_20200803-173837

Nikhil-Ladha:
Also, it would be better to add some margin around that button to keep it spottable in mobile view as it gets hidden within the code.

PR #593, opened 2020-08-06
adjusted scaling of blog images

### LINKED ISSUE #581: Images in blog not properly scaled
**Description**
The images in the blog cards are not properly scaled in [here](https://layer5.io/blog) . So, it needs to be fixed.

**Expected Result**
All the blog cards should be equally scaled & aligned in the [blog](https://layer5.io/blog) page.

**Screenshots**
L5

---
[Optional] **To Reproduce**
1. Go to 'https://layer5.io/blog'
2. Take the screenshot as the reference & go through the page
3. See error

Neilblaze:
**Description**
Fixed the images in the [blog](https://layer5.io/blog) cards which were not properly scaled.

This PR fixes #581 

### DIFF
--- a/_includes/collections/news.html
+++ b/_includes/collections/news.html
@@ -12,7 +12,7 @@
               max-width: 100%;
               max-height: 180px;
               width: auto;
-              height: auto;
+              height: 180px;
               text-align: center;
               margin-left: auto;
               margin-right: auto;
--- a/_includes/collections/posts.html
+++ b/_includes/collections/posts.html
@@ -12,7 +12,7 @@
               max-width: 100%;
               max-height: 180px;
               width: auto;
-              height: auto;
+              height: 180px;
               text-align: center;
               margin-left: auto;
               margin-right: auto;


Nikhil-Ladha:
Well, the height of the images is now good. But, if we could also make the width of them to be the same then it would be better.

Neilblaze:
> Well, the height of the images is now good. But, if we could also make the width of them to be the same then it would be better.

Well when I was making that PR, this came into my mind. But not every image will look good if I make that change since only one CSS class is applied to all of them for both `posts` & `news`

Nikhil-Ladha:
It will not look good if its stretched to do so, but i stead if using CSS it could be adjusted to fit the width then it will be ok, I think so. For a better idea if you van make the change and share 1 or 2 ss of the page, then it will be easier to consider.

Neilblaze:
So, at the moment, it looks -

Css unapplied

And after I did some tweaks to the css class, it now looks like - 

css applied


Now you say which is better :)

Nikhil-Ladha:
Haha😆! Cool..

PR #608, opened 2020-08-10
feat: ability to filter blogs by tags

Tanuj22:
**Description**

**Notes for Reviewers**


Tanuj22:
Yet to figure out what is causing the issue with the png on the scale-up effect of blog thumbnails

Nikhil-Ladha:
Everything looks good. Just a small update if the tags like `open-source` and some similar one's can be capitalized like others, it would look better.

src/sections/Navigation/utility/data.js
```diff
@@ -54,10 +54,6 @@ const Data = {
                 {
                     name: "Blog Grid",
                     path: "/blog-grid"
-                },
-                {
-                    name: "Blog List",
-                    path: "/blog-list"
```
leecalcote, line 60: Don't we want to keep this view as well?

Nikhil-Ladha: Here, we are using that view for the filtered blogs. But, I had a different thought for that view. And will discuss it in the meeting this Thursday. Then, we can decide on that part.

leecalcote: ok, @Nikhil-Ladha. Are we good to move forward in the meantime?

leecalcote: I have added to the meeting minutes so that we don't forget this discussion.

Nikhil-Ladha: I think we can hold it for a day. And, then we can move forward with what to do.

Tanuj22:
@leecalcote @Nikhil-Ladha commented out the `blog-list` page.

PR #618, opened 2020-08-13
Feature Image in Blog Page

Jashpatel1:
**Description**

This PR fixes #

**Notes for Reviewers**


Jashpatel1:
> Few changes:
> 
> 1. Please remove the  component from the blog pages
> 2. Apply these set of CSS to the class `page-header` to fix the issue of the image:
> 
> ```
>     height: auto;
>     margin: auto;
>     margin-top: 124px;
>     padding-top: 0px;
>     padding-bottom: 0px;
> ```
> 
> Few improvisation can be done like increasing the width and merging the padding and margins into one.

1. Can you please specify which component?
2. On using the above mentioned styles, the problem still persists.

Nikhil-Ladha:
> > Few changes:
> > 
> > 1. Please remove the  component from the blog pages
> > 2. Apply these set of CSS to the class `page-header` to fix the issue of the image:
> > 
> > ```
> >     height: auto;
> >     margin: auto;
> >     margin-top: 124px;
> >     padding-top: 0px;
> >     padding-bottom: 0px;
> > ```
> > 
> > 
> > Few improvisation can be done like increasing the width and merging the padding and margins into one.
> 
> 1. Can you please specify which component?
> 2. On using the above mentioned styles, the problem still persists.

src/ressouce/PageHeader/index.js
I will check if some other style was also needed. But, as far as I remember it fixes the issue.

Nikhil-Ladha:
> > Few changes:
> > 
> > 1. Please remove the  component from the blog pages
> > 2. Apply these set of CSS to the class `page-header` to fix the issue of the image:
> > 
> > ```
> >     height: auto;
> >     margin: auto;
> >     margin-top: 124px;
> >     padding-top: 0px;
> >     padding-bottom: 0px;
> > ```
> > 
> > 
> > Few improvisation can be done like increasing the width and merging the padding and margins into one.
> 
> 1. Can you please specify which component?
> 2. On using the above mentioned styles, the problem still persists.

I forgot to add `width:500px` to it. Add a fixed width and it will work.

Nikhil-Ladha:
> > > Few changes:
> > > 
> > > 1. Please remove the  component from the blog pages
> > > 2. Apply these set of CSS to the class `page-header` to fix the issue of the image:
> > > 
> > > ```
> > >     height: auto;
> > >     margin: auto;
> > >     margin-top: 124px;
> > >     padding-top: 0px;
> > >     padding-bottom: 0px;
> > > ```
> > > 
> > > 
> > > Few improvisation can be done like increasing the width and merging the padding and margins into one.
> > 
> > 
> > 
> > 1. Can you please specify which component?
> > 2. On using the above mentioned styles, the problem still persists.
> 
> I forgot to add `width:500px` to it. Add a fixed width and it will work.

leecalcote:
Looks better and better.

Nikhil-Ladha:
Isn't 90% too much? 
Well, i am on mobile right now. So, if it looks good with this update in desktop view, then it's fine.

Jashpatel1:
@Nikhil-Ladha Does 75% suits good?

Nikhil-Ladha:
Actually,  I meant that for mobile 90% looks good because the screen is small so if the image covers most of the screen it looks good. But, in case of bigger screens as we grow we may need to reduce the width as per screen size.
Also, there is a lot of white space on the top and bottom of the feature image, that can be reduced too.

Jashpatel1:
> Actually, I meant that for mobile 90% looks good because the screen is small so if the image covers most of the screen it looks good. But, in case of bigger screens as we grow we may need to reduce the width as per screen size.
> Also, there is a lot of white space on the top and bottom of the feature image, that can be reduced too.

For bigger screens I have added max-width: 500 px.

Nikhil-Ladha:
Oops, my eye!!!
Though, since you have added a better approach, I have one more thing to add, it is that the feature image looks good in that size, but we would want to have the `heading` or the name of the blog to be a bit wider, I mean relaxed and expanded rather than being compressed. So, since we have added the image using `gatsby-image` plugin the width of the image will not distort even if we increase the size of `class: page-header` to something around `1000px`, such that heading looks better. I have tried it for few of the blogs and it works fine for me. If you could confirm that from your end, and update that too.

PR #671, opened 2020-08-24
feat: choose between gird and list view in blogs

Tanuj22:
**Description**

**Notes for Reviewers**


Tanuj22:
@Nikhil-Ladha @leecalcote when filtering by tags in list view. Should it be in gird layout?

leecalcote:
Nice. Even though, the page location (the URL) is changing, the page is not fully reloading, correct? Just the main posts module?

It would be good to switch out the light red/pink highlight color with the light green from the Layer5 color scheme.

Tanuj22:
They are seperate pages so I think it is fully reloading. Will way a way to to to not make it fully reload

Tanuj22:
@Nikhil-Ladha what kind of icons were you thinking for list view and gird view?

Nikhil-Ladha:
Screenshot from 2020-08-12 18-15-24
Something like we have on this page, in the top right

Nikhil-Ladha:
The third icon is not needed only the first two will suffice.

Nikhil-Ladha:
Few things need to be updated:
1. Adding a tooltip on hover stating (Grid view/ List View)
2. A background color while hovering, as there was previously.
3. If the size of icons can be increased by 2 px or so it would look better.
4. A bit of design modification like a single border taking up both icons also a vertical line separation kind of thing between both of them.
5. Highlight the icon w.r.t the view the page is in currently.
6. Also, we don't want the icons when we are seeing the individual blog posts.
7. Take care of the responsiveness of the icons, it goes to the bottom of the blogs in the mobile view.

I know it may seem like too much, but I feel the end product will be something you to look at!

Nikhil-Ladha (changes requested):
Please address the above points

Tanuj22:
@Nikhil-Ladha I have made some changes. I think point 4 and 5 are left. As Lee pointed out the the page shouldn't full reload, so should I only create a page say `/blog` and switch between gird and list view keeping that URL?

Nikhil-Ladha:
> @Nikhil-Ladha I have made some changes. I think point 4 and 5 are left. As Lee pointed out the the page shouldn't full reload, so should I only create a page say `/blog` and switch between gird and list view keeping that URL?

Yeah, you can just do a conditional rendering of the components. That would fix it.

Tanuj22:
@Nikhil-Ladha please have a look

Nikhil-Ladha:
> @Nikhil-Ladha please have a look

Nice updates here. Few minor changes:
Initially, the tooltip doesn't have a background color. But after clicking the background color is coming up. 
See the screenshots below for the tooltips:
Screenshot from 2020-08-29 20-02-18
Screenshot from 2020-08-29 20-02-56

Also, I think you are yet to implement point 4 that I mentioned above.
Otherwise, it looks much better!!!

Tanuj22:
@Nikhil-Ladha fixed the isse with tooltip. Can someone less look at point 4 as I am not able to implement it nicely?

Tanuj22:
There seems to an issue with netlify - https://github.com/netlify/cli/issues/1163

Jashpatel1:
> @Nikhil-Ladha fixed the isse with tooltip. Can someone less look at point 4 as I am not able to implement it nicely?

I will correct it.

Tanuj22:
Netlify working now

Nikhil-Ladha:
Sorry, for asking it so late.
I think we don't need to provide this view option in the mobile view, as the grid is better and works fine for the mobile view. The list view doesn't look so soothing in the mobile view.
If you could just add a breakpoint update for the above thing. Then, we will be good to go. The rest all looks great!

Nikhil-Ladha:
Look's good. @Jashpatel1 will you amke a seperate PR for the icons bordering?
Let's merge this one now. Good work @Tanuj22 👌🏼

Nikhil-Ladha (approved):

PR #754, opened 2020-09-10
Blog List View Changes

Jashpatel1:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/src/components/blog-view-tooltip.js
+++ b/src/components/blog-view-tooltip.js
@@ -1,9 +1,9 @@
-import React from "react"
-import { Row } from "../reusecore/Layout"
-import { AiOutlineUnorderedList } from "react-icons/ai"
-import { BsFillGrid3X3GapFill } from "react-icons/bs"
-import ReactTooltip from "react-tooltip"
-import styled from "styled-components"
+import React from "react";
+import { Row } from "../reusecore/Layout";
+import { AiOutlineUnorderedList } from "react-icons/ai";
+import { BsFillGrid3X3GapFill } from "react-icons/bs";
+import ReactTooltip from "react-tooltip";
+import styled from "styled-components";
 
 export const ToolTipWrapper = styled.div`
       @media screen and (max-width: 576px) {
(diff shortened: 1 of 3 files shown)


Tanuj22:
@Jashpatel1 can you format the code my running `npm run lint`

Tanuj22:
The tooltips are looking different. Also the icons have more space at the bottom?
ezgif-7-3ba0943b468f

Jashpatel1:
> The tooltips are looking different. Also the icons have more space at the bottom?
> ezgif-7-3ba0943b468f

I havent changed anything to the icons or spacing even, just added a border and changed colours a bit.

Nikhil-Ladha:
The divider in the middle looks a bit odd, like there is a gap in the bottom. 
Also, if the box can be filled with color on selection and hover. It will make it look better.

leecalcote:
Gents, I will submit to you that a border here is a misstep - overkill.  A `|` might do the trick or some other softer approach.

Nikhil-Ladha:
Actually, yeah. Here, it looks like that but the picture I had in mind was something different. So, I will try if I can do something to create it visually and then take some feedback on that.

leecalcote (changes requested):
https://github.com/layer5io/layer5/pull/754#issuecomment-690618566

leecalcote:
I don't think we got a chance to review this on yesterday's call, @Jashpatel1, but I'm hoping that you have the feedback you need to move forward.

Nikhil-Ladha:
Oops, if you don't get anything from me by tonight please do make the changes Lee mentioned. Otherwise, can you please wait until tomorrow.

Nikhil-Ladha:
So, here it is what I wanted to say in the first place
Screenshot from 2020-09-15 19-14-10
Screenshot from 2020-09-15 19-12-13

Does this look better?
@leecalcote @Jashpatel1 @Tanuj22 ?

Tanuj22:
Yeah. This does look nice!

Nikhil-Ladha:
Then, I think @Jashpatel1 you can implement this, preferably the first one I guess.

leecalcote:
It'll be helpful to see this new style in-context of the rest of the page.

leecalcote:
@Jashpatel1 I hate to be the bearer of bad news. The coloring is ... well, off. I don't know that I have better direction to give here other than to suggest that this isn't quite the design that we would like to have. We can accept this PR and move on for now with the intent return to style these buttons later.

leecalcote (dismissed):

Nikhil-Ladha:
Just a small change would be good, on hovering the colors should just revert and the gray should come up. Like, we have for the `Grid` icon when the page loads, it should remain as it is on hovering, and the other one should become the same.

Jashpatel1:
> Just a small change would be good, on hovering the colors should just revert and the gray should come up. Like, we have for the `Grid` icon when the page loads, it should remain as it is on hovering, and the other one should become the same.

I didn't quite understand what you are trying to convey. Can you please elaborate?

Nikhil-Ladha:
Basically, what I meant was to remove the gray😅.
And ,on hovering it should look similar to what it looks when the button is `active`

Nikhil-Ladha (approved):

PR #822, opened 2020-09-26
[NG] Add and move Layer5 logos

leecalcote:
**Description**
Layer5 logo refresh for NG site


leecalcote:
I suspect that I have a file reference out of place, but am not sure. I'm receiving a number of these warnings on build:

```
warn You can't use childImageSharp together with feature-image.svg — use publicURL instead. The childImageSharp
portion of the query in this file will return null:
```

Jashpatel1:
> I suspect that I have a file reference out of place, but am not sure. I'm receiving a number of these warnings on build:
> 
> ```
> warn You can't use childImageSharp together with feature-image.svg — use publicURL instead. The childImageSharp
> portion of the query in this file will return null:
> 
@leecalcote Its a common warning we are looking forward to resolve.

leecalcote:
Oh, I see. Does visiting /blogs crash the local non server?

Jashpatel1:
> Oh, I see. Does visiting /blogs crash the local non server?

Yes, we need to create redirect paths from /blogs to /blog .

Tanuj22:
> I suspect that I have a file reference out of place, but am not sure. I'm receiving a number of these warnings on build:
> 
> ```
> warn You can't use childImageSharp together with feature-image.svg — use publicURL instead. The childImageSharp
> portion of the query in this file will return null:
> ```

This warning is due to the use of svgs with Gatsby Image. It is a workaround that was suggested in the gatsby docs. But going so introduced this warning.

Tanuj22 (approved):

PR #826, opened 2020-09-26
feat: careers, programs and internships page made dynamic

Tanuj22:
**Description**

This PR fixes #

**Notes for Reviewers**


Nikhil-Ladha:
We have more programs that needs to be added, right?
Also, the Full time opportunities section is missing only the header is there.

Tanuj22:
> We have more programs that needs to be added, right?
> Also, the Full time opportunities section is missing only the header is there.

Yes. For both of them individual issues can be created.

Nikhil-Ladha:
Means, the data has not been migrated yet?

Tanuj22:
Yeah

Tanuj22:
Can you give an example of the text color where is not correct?

Nikhil-Ladha:
> Can you give an example of the text color where is not correct?

In the programs page, the text should be black, but in some places its still gray.

Nikhil-Ladha:
Actually, not only there many places the text is gray, but it should be black, right? Like the Digital Marketing Internship page.

Tanuj22:
Yes, it should be black. There might be some global style messing this up. Checking now

> Actually, not only there many places the text is gray, but it should be black, right? Like the Digital Marketing Internship page.

Nikhil-Ladha:
It should be `Home > Careers`
Screenshot_20200929-203915__01

This part can be removed for this page and other consequent pages.
Screenshot_20200929-203857__01__01

Tanuj22:
@Nikhil-Ladha Please take a look

Nikhil-Ladha (approved):
Looks good ✅

Issue #884, opened 2020-10-01
[Programs] Add Layer5 Community Hacktoberfest logo to program page


leecalcote:
**Current Behavior**
The Layer5 community have a special Hacktoberfest graphic to share on this page - https://layer5.io/programs/hacktoberfest

**Desired Behavior**
Add the graphic below to the Hacktoberfest page. Replace the existing generic graphic with this one.
![Layer5-Hacktoberfest](https://user-images.githubusercontent.com/7570704/94851172-b1336500-03ed-11eb-917c-9bdb67ba8684.png)

---
**Contributor Resources**
The `layer5` repo contains two websites. The current generation and the next-generation of the layer5.io site.

If the `layer5-ng` label is absent on this issue, then this issue pertains to the current generation of the layer5.io website, which uses Jekyll and GitHub Pages. Site content is found under the [`master` branch](https://github.com/layer5io/layer5/tree/master).
- See [`layer5` contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)

If the `layer5-ng` label is present on this issue, then this issue pertains to the next-generation of the layer5.io website, which uses Gatsby, Strapi, and GitHub Pages. Site content is found under the [`layer5-ng` branch](https://github.com/layer5io/layer5/tree/layer5-ng).
- See [`layer5-ng` contributing instructions](https://github.com/layer5io/layer5/blob/layer5-ng/CONTRIBUTING.md)

bartkalanski:
I’m away for a little birthday celebration but will try and get it done by Sunday if that’s okay ?

leecalcote:
@bartkalanski thanks much. Yes, sounds real good. The next Layer5 Hacktoberfest meetup is on Thursday, October 8th.

PR #906, opened 2020-10-03
Replaced generic graphic with Hacktoberfest graphic and added alt text

bartkalanski:
**Description**

Issue #884 "[Programs] Add Layer5 Community Hacktoberfest logo to program page"

This PR replaces generic graphic with Hacktoberfest graphic and adds alt text

Signed-off-by Bart Kalanski bkalanski@gmail.com
**Notes for Reviewers**


collections/_programs/hacktoberfest.html
```diff
@@ -8,7 +8,7 @@
 permalink: /programs/hacktoberfest
 excerpt: "Layer5 - Hacktoberfest"
 ---
-<img src="/assets/images/programs/hacktoberfest2020/hacktober.svg" class="image-center" />
+<img src="https://user-images.githubusercontent.com/7570704/94851172-b1336500-03ed-11eb-917c-9bdb67ba8684.png" alt="Hacktoberfest and Layer5 logo" class="image-center" />
```
Jashpatel1, line 11: Can you download the graphic in assets folder and then add it? Also the height to width ratio in current preview looks disturbed.

bartkalanski: No problem, that makes more sense!

bartkalanski: The width for the image is set at 40%, I set it to 60% in chrome devtools and it does seem to look better. Would you say that's the best way of improving the issue that you mentioned ?

Jashpatel1: I think @leecalcote and @chandrashritii can look into if we could have a wider graphic.

Jashpatel1 (dismissed):
Nice work. Thanks @bartkalanski 
Can something be done with the image @leecalcote @chandrashritii ?

bartkalanski:
> Nice work. Thanks @bartkalanski
> Can something be done with the image @leecalcote @chandrashritii ?

Thank you! Pleasure as always :)

Nikhil-Ladha:
Also, in mobile view the image size can be increased a bit.

leecalcote:
If the image is moved into the left column, this will solve the issue of too much whitespace (the image being much taller than it is wider). It will also solve part of the problem of the questionnaire feeling so tall vs the text on the left.

Screen Shot 2020-10-04 at 11 22 59 AM

@bartkalanski, would you mind relocating?

bartkalanski:
> If the image is moved into the left column, this will solve the issue of too much whitespace (the image being much taller than it is wider). It will also solve part of the problem of the questionnaire feeling so tall vs the text on the left.
> 
> Screen Shot 2020-10-04 at 11 22 59 AM
> 
> @bartkalanski, would you mind relocating?

I'm happy to have a go at this, would you like to me to move the image into the left column only ? Or the hacktoberfest heading as well ?

Jashpatel1:
> I'm happy to have a go at this, would you like to me to move the image into the left column only ? Or the hacktoberfest heading as well ?

The graphic would be enough I guess.

Nikhil-Ladha:
I think the second heading `Hactoberfest 2020 with Layer5` can be removed or switched with the outer title and remove this one.
Also, if you could adjust the height of the iframe to be same as that of the left column having the paragraph.

Nikhil-Ladha:
I don't see the iframe height adjusted to be same as that of the left column.
Also, the image size needs to be increased a bit for mobile view.

Nikhil-Ladha:
Also, please resolve the merge conflict too

leecalcote:
@Nikhil-Ladha we might be able to forego the iframe height update. I made the iframe this tall in order to fully display the questionnaire without the need for scrolling (so that the user doesn't have a scrollbar inside of another scrollbar). The text in the lefthand column is vertically-aligned to the top. 

If you consider that it's a better experience to have the iframe height adjust on the form, though, no sweat, that'll work as well.

leecalcote:
@Nikhil-Ladha good call about the mobile view. We have an issue with the google form spilling off the page, too.

Nikhil-Ladha:
> @Nikhil-Ladha good call about the mobile view. We have an issue with the google form spilling off the page, too.

Already opened an issue for that...and it's been worked upon by someone.🙂

Nikhil-Ladha:
> @Nikhil-Ladha we might be able to forego the iframe height update. I made the iframe this tall in order to fully display the questionnaire without the need for scrolling (so that the user doesn't have a scrollbar inside of another scrollbar). The text in the lefthand column is vertically-aligned to the top.
> 
> If you consider that it's a better experience to have the iframe height adjust on the form, though, no sweat, that'll work as well.

Just to omit the white space beneath the text. If that's intentional we may skip that. But I think it would look a bit odd.

leecalcote:
Screen Shot 2020-10-05 at 11 10 13 AM

This whitespace, right? Given that the form is embedded in the div, unless we cut the form in half, we won't be able to be rid of the whitespace. Cutting the form in half will have the effect of removing the whitespace and making the user scroll on the form.

Sounds like your thinking is that this might be a better user experience. Let's go with your suggestion and reduce the height on the iframe. 👍

bartkalanski:
> I don't see the iframe height adjusted to be same as that of the left column.
> Also, the image size needs to be increased a bit for mobile view.

My bad, I assumed you meant for it to be inline with the paragraph. Do you have any advice on how I could reduce the size of the iframe ?

bartkalanski:
> Also, please resolve the merge conflict too

All done, but I checked the new preview and it's looking wrong

bartkalanski:
> Screen Shot 2020-10-05 at 11 10 13 AM
> 
> This whitespace, right? Given that the form is embedded in the div, unless we cut the form in half, we won't be able to be rid of the whitespace. Cutting the form in half will have the effect of removing the whitespace and making the user scroll on the form.
> 
> Sounds like your thinking is that this might be a better user experience. Let's go with your suggestion and reduce the height on the iframe. 👍

I should be able to achieve this by setting height to 1405 and adding scrolling="yes" attribute

Jashpatel1 (approved):
As the spilling in mobile view is worked upon as a different issue, I guess this PR is complete.

leecalcote (approved):
Fantastic. :)

bartkalanski:
I noticed I must’ve put the old hacktoberfest  graphic source code in the latest commit, which means the old graphic is showing instead of the new one, do you mind if I quickly fix that ?

leecalcote:
Hi Bart, not at all. Please do.

PR #921, opened 2020-10-05
feat : add pagination plugin

### LINKED ISSUE #521: [NG] Add pagination plugin
**Desired Behavior**
Add and configure [gatsby plugin for pagination](https://www.gatsbyjs.org/packages/gatsby-awesome-pagination/) in[ layer5-ng](https://github.com/layer5io/layer5/tree/layer5-ng). Related to #481

Tanuj22:
**Description**

This PR fixes #521

**Notes for Reviewers**

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -7,23 +7,23 @@
 
 const path = require(`path`);
 const slugify = require("./src/utils/slugify");
+const { paginate } = require("gatsby-awesome-pagination");
 
-// You can delete this file if you're not using it
 // Replacing '/' would result in empty string which is invalid
-const replacePath = path => (path === `/` ? path : path.replace(/\/$/, ``))
+const replacePath = path => (path === `/` ? path : path.replace(/\/$/, ``));
 // Implement the Gatsby API “onCreatePage”. This is
 // called after every page is created.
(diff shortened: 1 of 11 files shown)


gatsby-node.js
```diff
@@ -124,6 +127,14 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
 
   const members = res.data.allMembers.nodes;
 
+  paginate({
+    createPage,
+    items: blogs,
+    itemsPerPage: 8,
+    pathPrefix: `/blog`,
+    component: blogViewTemplate
```
leecalcote, line 135: Will we need to update this code in the future to accommodate news, workshops, and other collections? If so, is there a way to make this less fragile?

Tanuj22: The paginate function would need to be called for each of the collections in that case

leecalcote: Ok, a centrally-controlled configuration to enable pagination on specific collections. This makes sense.

Tanuj22:
Added pagination for blogs and when filtered by tags. Things like jumping to a particular page, the last page, etc. can be done. I can come back to that later.

leecalcote (approved):

PR #992, opened 2020-10-21
redirection of event link to new tab

vinayaksh42:
Signed-off-by: Vinayak Sharma <vinayaksh42@gmail.com>

**Description**
redirection of event link to a new tab

### DIFF
--- a/src/components/EventCard/Card.style.js
+++ b/src/components/EventCard/Card.style.js
@@ -95,8 +95,8 @@ export const CardWrapper = styled.div`
     }
     .readmore-btn {
        color: rgba(0,0,0,0.35);
-       display: flex;
-       align-items: center;
+       align-items: right;
+       float: right;
        transition: all 0.3s linear;
        svg{
            margin-left: 0px;
--- a/src/components/EventCard/index.js
+++ b/src/components/EventCard/index.js
@@ -10,7 +10,7 @@ const Card = ({ frontmatter }) => {
         <CardWrapper fixed={!!frontmatter.abstract}>
             <div className="post-block">
                 <div className="post-thumb-block">
-                    <a className="anchor" href={frontmatter.eurl}>
+                    <a className="anchor" target="_blank" href={frontmatter.eurl}>
                         <Image
                             {...frontmatter.thumbnail}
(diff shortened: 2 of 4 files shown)


Nikhil-Ladha:
I guess, there is some security issues with using the target attribute directly....please check the details of the test failure and fix it accordingly, so that we can move ahead.

Nikhil-Ladha:
Also, sorry that I didn't check these things in the previous PR but two more changes need to be done
1. The top heading `Events` with the page path is not needed.
2. And, in the cards the `see more` button and the `icon` both perform the same thing, so better to remove the `see more` button in this page.

Nikhil-Ladha (changes requested):
^^

Nikhil-Ladha (changes requested):
- Fix warnings about unused imports in line:2
- The `Icon` to redirect the user to a different page should be on the same line as that of date.
- For the meetups, as we don't have a date field simply just remove the icon in that card.
- Change the heading `Events`, `Meetups` to use `<h1>` tags, and remove the margin, increase the padding.

vinayaksh42:
> * Fix warnings about unused imports in line:2
> * The `Icon` to redirect the user to a different page should be on the same line as that of the date.
> * For the meetups, as we don't have a date field simply just remove the icon in that card.
> * Change the heading `Events`, `Meetups` to use `<h1>` tags, and remove the margin, increase the padding.

The meetup card and the event card are utilizing the same card so to remove the date field for the meetup cards would mean to make a separate card component for them. Also in the meetup cards, anything related to dates is not visible

Nikhil-Ladha:
> The meetup card and the event card are utilizing the same card so to remove the date field for the meetup cards would mean to make a separate card component for them. Also in the meetup cards, anything related to dates is not visible

I didn't ask you to remove the date field. Simply, check if the date is there only then show the icon otherwise don't.

Nikhil-Ladha:
Also, while you do so change the top padding to 100px from 80px for the heading.

Nikhil-Ladha (approved):
Thank you @vinayaksh42 👍🏼

PR #1178, opened 2020-11-22
Updates on blog and workshops page on Layer5-ng

Jashpatel1:
Signed-off-by: Jashpatel1 <patel.5@iitj.ac.in>

**Description**

This PR makes updates on different pages on Layer5-ng.

**Notes for Reviewers**


src/components/Card/index.js
```diff
@@ -10,44 +10,44 @@ const Card = ({ frontmatter, fields }) => {
...
-                            see more <IoIosArrowRoundForward />
-                        </Link>
-                        {frontmatter.eurl && (
-                            <a className="readmore-btn" href={frontmatter.eurl}>
-                                <BiLinkExternal />
-                            </a>
-                        )}
+                    <div className="post-content-block">
+                        <h2 className="post-title">
+                            <Link className="anchor" to={fields.slug}>
```
Tanuj22, line 23: There is no need for a `Link` tag above these when it's present on the entire card. Do this for other places

src/components/Card/index.js
```diff
@@ -10,44 +10,44 @@ const Card = ({ frontmatter, fields }) => {
...
+                                {frontmatter.title}
+                            </Link>
+                        </h2>
+                        <div className="post-meta-block">
+                            {frontmatter.author && <span>By: {frontmatter.author}</span>}
+                            {frontmatter.date && (
+                                <>
+                                    <span className="divider">/</span>
```
Tanuj22, line 31: Remove divider as discussed. The date can then be left aligned

src/templates/blog.js
```diff
@@ -26,7 +26,7 @@ export const query = graphql`
                 id
                 frontmatter {
                     title
-                    date(formatString: "Do MMMM YYYY")
+                    date(formatString: "MMMM Do, YYYY")
```
Tanuj22, line 29: Do this for news and any other places dates are used

src/components/Card/index.js
```diff
@@ -10,44 +10,44 @@ const Card = ({ frontmatter, fields }) => {
...
-                            </a>
-                        )}
+                    <div className="post-content-block">
+                        <h2 className="post-title">
+                            <Link className="anchor" to={fields.slug}>
+                                {frontmatter.title}
+                            </Link>
+                        </h2>
+                        <div className="post-meta-block">
+                            {frontmatter.author && <span>By: {frontmatter.author}</span>}
```
Tanuj22, line 28: The `By` was supposed to be removed right?

Tanuj22 (changes requested):
The arrow animation on `see more` should occur as soon as we hover on the card now and not when we go near it

src/components/Card/index.js
Jashpatel1: It was discussed to have the entire card clickable.

src/components/Card/index.js
Jashpatel1: I didn't find it in the doc. I will do that if required.

src/components/Card/index.js
Tanuj22: So just put a link tag on the entire card instead of individual places

leecalcote:
You will want to double check on the blog post card in grid view, specifically the post date and the post author don’t seem to be on the same line (they are vertically misaligned).

leecalcote:
Nearest I can tell him I’m on the phone, the workshops card is much improved.

Nikhil-Ladha:
One thing, should we add something like `Read more` with a `down arrow` when the card is not expanded and then change it to just `^` arrow on expansion?
It would not be specially focused or designed button just like a text element or so?
For UX?

Jashpatel1:
@Nikhil-Ladha added the arrow

Nikhil-Ladha:
Why is the arrow pointed diagonally?
And also the text `Read more` is missing.

Jashpatel1:
> Why is the arrow pointed diagonally?
> And also the text `Read more` is missing.
> One more thing, why ae we adding a `Green` border on the top of each card in the workshops page?

I thought as the card is expanding in both directions, so a diagonal arrow.
I will add the text.
I don't know about the border part. Will look into it.

Nikhil-Ladha:
> I thought as the card is expanding in both directions, so a diagonal arrow.

Hmmm.....ypu are right at your point. But, not sure about this.
What do others have to say in this?
@Tanuj22 @leecalcote

Tanuj22:
> > I thought as the card is expanding in both directions, so a diagonal arrow.
> 
> Hmmm.....ypu are right at your point. But, not sure about this.
> What do others have to say in this?
> @Tanuj22 @leecalcote

It makes sense logically but it's feeling kind of out of place. I would go with the downward pointing arrow 😅

Nikhil-Ladha (approved):
Nice! :100:

Tanuj22 (approved):
😆

PR #1218, opened 2020-12-04
megamenu fixes

vineethvanga18:

**Description**

- fetching of blogs dynamically

- fixes and improvements in animations

- minor naming and UI fixes

**Notes for Reviewers**

### DIFF
--- a/src/sections/Navigation/index.js
+++ b/src/sections/Navigation/index.js
@@ -17,7 +17,7 @@ const Navigation = () => {
             query  {
                 Projects: allMdx(
                     sort: { fields: [frontmatter___date], order: DESC }
-                    filter: { fields: { collection: { eq: "projects" } } }
+                    filter: { fields: { collection: { eq: "events" } } }
                     limit: 2
                 ) {
                     nodes {
@@ -25,6 +25,11 @@ const Navigation = () => {
                         frontmatter {
                             title
                             thumbnail{
+                                childImageSharp{
+                                    fluid(maxHeight: 1000){
+                                        ...GatsbyImageSharpFluid_withWebp
+                                    }
+                                }
                                 publicURL
                             }
                         }
@@ -43,6 +48,11 @@ const Navigation = () => {
                         frontmatter {
                             title
                             thumbnail{
+                                childImageSharp{
+                                    fluid(maxWidth: 1000){
+                                        ...GatsbyImageSharpFluid_withWebp
+                                    }
+                                }
                                 publicURL
                             }
                         }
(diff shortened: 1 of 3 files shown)


shivaylamba (commented):
image
@vineethvanga18 the images are a bit stretched. I think we can make the mega menu slightly smaller as well.

vineethvanga18:
yes @shivaylamba, One solution would be to fetch the correct sized image through graphql

leecalcote:
Thank you, @vineethvanga18. Two other issues to address:

1. When the mouse approaches the menu from underneath, the menu is activated well in advance of the mouse reaching the menu text.
1. The exit animation is not invoked when the mouse leaves from the top of the menu.

Dec-04-2020 17-56-30

kishore007k:
Nice Work @Vineet-Sharma29 
One small change, It would be nice if we display the upcoming events/workshops in the community instead of displaying the most recent ones.
What are your thoughts @leecalcote Sir?

Nikhil-Ladha:
> Nice Work @Vineet-Sharma29
> One small change, It would be nice if we display the upcoming events/workshops in the community instead of displaying the most recent ones.
> What are your thoughts @leecalcote Sir?

+1. Good point @kishore007k

src/sections/Navigation/index.js
```diff
@@ -11,6 +12,102 @@ import ScrollspyMenu from "./utility/ScrollspyMenu.js";
...
+                    sort: { fields: [frontmatter___date], order: DESC }
+                    filter: { fields: { collection: { eq: "projects" } } }
+                    limit: 2
+                ) {
+                    nodes {
+                        id
+                        frontmatter {
+                            title
+                            thumbnail{
+                                publicURL
```
Tanuj22, line 28: Here childImageSharp should be used like (reference can be taken from other queries)

src/sections/Navigation/utility/ScrollspyMenu.js
```diff
@@ -51,12 +65,12 @@ const ScrollspyMenu = ({ menuItems, ...props }) => {
                             </div>
                             <div className="nav-display">
                                 <div className="nav-div">
-                                    <a href={active.div1.path}><img src={active.div1.src} className="nav-img" /></a>
-                                    <div> {active.div1.descr} </div>
+                                    <Link href={blogData[active.name].nodes[0].fields.slug}><img src={blogData[active.name].nodes[0].frontmatter.thumbnail.publicURL} className="nav-img" /></Link>
```
Tanuj22, line 68: After using childImageSharp. `Image` component can be used

vineethvanga18: What are the available image sizes @Tanuj22

Tanuj22: Any size will work

leecalcote (approved):
We do need to reduce the overall size of this menu, but in its current form it is pretty **_awesome_**, @vineethvanga18.  🥇

PR #1236, opened 2020-12-08
Replaced carousel library

OmkarPh:
Signed-off-by: OmkarPh <omkarphansopkar@gmail.com>

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6724,6 +6724,14 @@
       "resolved": "https://registry.npmjs.org/dom-walk/-/dom-walk-0.1.2.tgz",
       "integrity": "sha512-6QvTW9mrGeIegrFXdtQi9pk7O/nSK6lSdXW2eqUspN5LWD7UTji2Fqw5V2YLjBpHEoU9Xl/eUWNpDeZvoyOv2w=="
     },
+    "dom7": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/dom7/-/dom7-3.0.0.tgz",
+      "integrity": "sha512-oNlcUdHsC4zb7Msx7JN3K0Nro1dzJ48knvBOnDPKJ2GV9wl1i5vydJZUSyOfrkKFDZEud/jBsTk92S/VGSAe/g==",
+      "requires": {
+        "ssr-window": "^3.0.0-alpha.1"
+      }
+    },
     "domain-browser": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/domain-browser/-/domain-browser-1.2.0.tgz",
@@ -7095,11 +7103,6 @@
           }
         }
       }
-    },
-    "enquire.js": {
-      "version": "2.1.6",
-      "resolved": "https://registry.npmjs.org/enquire.js/-/enquire.js-2.1.6.tgz",
-      "integrity": "sha1-PoeAybi4NQhMP2DhZtvDwqPImBQ="
     },
     "enquirer": {
       "version": "2.3.6",
@@ -12523,14 +12526,6 @@
       "resolved": "https://registry.npmjs.org/json-stringify-safe/-/json-stringify-safe-5.0.1.tgz",
       "integrity": "sha1-Epai1Y/UXxmg9s4B1lcB4sc1tus="
     },
-    "json2mq": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/json2mq/-/json2mq-0.2.0.tgz",
-      "integrity": "sha1-tje9O6nqvhIsg+lyBIOusQ0skEo=",
-      "requires": {
-        "string-convert": "^0.2.0"
-      }
-    },
     "json3": {
       "version": "3.3.3",
       "resolved": "https://registry.npmjs.org/json3/-/json3-3.3.3.tgz",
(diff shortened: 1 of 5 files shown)


OmkarPh:
Hi @Jashpatel1 
I realized that react-slick is already in use by component slick.slider.js in reusecore

In the preview, I've shown carousel using both react slick and swiper JS(2nd one)

React slick ( 1st one ) looks better to me.

I tried shifting arrows in Swiper JS outward but it has a overflow hidden property (necessary) which forces to conjest arrows.

Jashpatel1:
> Hi @Jashpatel1
> I realized that react-slick is already in use by component slick.slider.js in reusecore
> 
> In the preview, I've shown carousel using both react slick and swiper JS(2nd one)
> 
> React slick ( 1st one ) looks better to me.
> 
> I tried shifting arrows in Swiper JS outward but it has a overflow hidden property (necessary) which forces to conjest arrows.

@leecalcote @Nikhil-Ladha @Tanuj22 ^^

Tanuj22:
I am not seeing any difference in both of them :sweat_smile:. If you say so we can go with the 1st one. Though the arrows are not looking nice currently. Also, the mobile view needs to be taken care of.

OmkarPh:
I prefer 1st one not because of styling, it can be manipulated as we want (even like 2nd ones)

But in 2nd one the arrows and corner cards are so close and pose problems in wider cards

Nikhil-Ladha:
Okay, so I have a different opinion here. In mobile, the second is better because of the feel/animation while swiping the cards.

>But in 2nd one the arrows and corner cards are so close and pose problems in wider corner cards

I think there should be a way to manipulate that as well, if not directly then you can refer to the class that is being used by inspecting and then add CSS for it. That will work.

OmkarPh:
> But in 2nd one the arrows and corner cards are so close and pose problems in wider corner cards  

There were 2 ways to solve this:

1. Provide a margin for arrows
Coz of this, arrows eventually become invisible partially because of hidden overflow.

2. Reduce size of cards wrapper
 This blog wrapper is overflowed by default in component so modifying its width is useless.

I tried many other things but unsuccessful ☹️

leecalcote:
I agree that the 1st one has a smoother, more desirable swiping animation and that we should try to retain this animation whether the 1st or 2nd one is used.

OmkarPh:
Should i make the arrow of both like second one?
Plain blue or layer5's green

Nikhil-Ladha:
Layer5 green

OmkarPh:
Can you tell me how to view this on mobile within same network ( not available on PC's ip now ) ?
I usually test with dev tools but for proper swipe feeling, wanted to test on my mobile

OmkarPh:
Note - I've used window object for toggling arrows and dots, so reload after changing resolution in dev tools :)

Jashpatel1 (commented):
The first one looks good to me.

leecalcote:
Regarding the first implementation, note that the righthand arrow flickers and half hides itself on hover.

Nikhil-Ladha:
I think we would want to use the colors in the following way?
Screenshot from 2020-12-10 11-39-41

OmkarPh:
> I think we would want to use the colors in the following way?

@Tanuj22 @Jashpatel1 @leecalcote ^^

Tanuj22:
> > I think we would want to use the colors in the following way?
> 
> @Tanuj22 @Jashpatel1 @leecalcote ^^

I agree with @Nikhil-Ladha . Rest looks great!. Do remember to take care of the mobile view :)

OmkarPh:
Is current disabled color ok?
image

Nikhil-Ladha:
Not sure if we want to keep the dots or not, is there any way to remove them?
Also, hide the arrow if it's disabled.

OmkarPh:
> Not sure if we want to keep the dots or not, is there any way to remove them?
> Also, hide the arrow if it's disabled.


In mobile view, user would generally prefer a swiping so I disabled arrows there.
Without dots, users might not understand if this is a carousel at all and they can swipe too. So kept dots only for mobile.

Nikhil-Ladha:
> > Not sure if we want to keep the dots or not, is there any way to remove them?
> > Also, hide the arrow if it's disabled.
> 
> In mobile view, user would generally prefer swiping so I disabled arrows there.
> Without dots, users might not understand if this is a carousel at all and they can swipe too. So kept dots only for mobile.

For mobile view, we might keep that. But not for desktop view.
Also, please check the alignment of carousel in mobile view. It's not center aligned.

Jashpatel1 (approved):
Looks good. Thanks @OmkarPh for this.

Tanuj22:
The mobile view is left ...

OmkarPh:
> The mobile view is left ...

I fixed that, the cards are centred and take up maximum possible space horizontally now.

Tanuj22:
> > The mobile view is left ...
> 
> I fixed that, the cards are centred and take up maximum possible space horizontally now.

Why I am seeing this :thinking: 
Screenshot from 2020-12-11 14-20-45

Tanuj22:
Nevermind it was due to directly switching to mobile from desktop view without reloading :sweat_smile:

OmkarPh:
Also this tablet view is taken care of:
Screenshot_2020-12-11-14-21-28-23

OmkarPh:
> Nevermind it was due to directly switching to mobile from desktop view without reloading

Ya that is a problem due to window obj😅
Do we have anything else that can be used to get innerwidth in gatsby?

Tanuj22:
> > Nevermind it was due to directly switching to mobile from desktop view without reloading
> 
> Ya that is a problem due to window obj
> Do we have anything else that can be used to get innerwidth in gatsby?

Not at the top of my head. Will let you know if I find anything

OmkarPh:
I found this, https://reedbarger.com/how-to-create-a-usewindowsize-react-hook/

PR #1281, opened 2020-12-18
Updated the careers page design

Nikhil-Ladha:

**Description**

This PR updates the careers page design.

**Note for reviewers**
Hope that it looks good 😁 


Jashpatel1 (changes requested):
Great change. Just a few things:
Lee's image of CCOSS looks stretched horizontally.
The transitions(shadows) come up a bit late I guess.
I didn't understand why the size of full-time opportunities was reduced horizontally if only 3 were to be fit in a single row.
Button Link in join the community section is incorrect.
Unused imports

Nikhil-Ladha:
> The transitions(shadows) come up a bit late I guess.
It's merely 0.3s. 

> I didn't understand why the size of full-time opportunities was reduced horizontally if only 3 were to be fit in a single row.

The size is not reduced the card is made in that way, that so as to adapt to the size of content within it.
4 looks too much cozy, so kept them as 3 as well.

> Button Link in join the community section is incorrect.

Ahh, yes forgot to update that.

Nikhil-Ladha:
The unused import for path, was there from before. I think few components still, pass it but it's commented in the actual code.

src/sections/Careers/index.js
```diff
@@ -12,54 +13,60 @@ const CareersPage = () => {
...
+                        </p>
+                        <p>
+                            We believe that all contributors should expect and be part of a safe and
+                            friendly environment for constructive contribution. We can more
+                            effectively and successfully compare and challenge different ideas to find
+                            the best solutions for advancement, while building the size, diversity,
+                            and strength of our community.
+                        </p>
+                    </div>
+                    <div className="opensource-section-img">
```
Tanuj22, line 57: Should we make a component out of this section image as the prop so we can use it in other places also?

Tanuj22 (commented):
Rest LGTM!

src/sections/Careers/index.js
Nikhil-Ladha: I don't think that's needed. The usage of images varies a lot. And creating a component for it, won't be beneficial at all.

Tanuj22: Yeah makes sense

Jashpatel1 (approved):
Looks good. @Tanuj22 should we proceed with merging or you fixing the failing tests?

Nikhil-Ladha:
> Looks good. @Tanuj22 should we proceed with merging or you fixing the failing tests?

@Jashpatel1 fixed now.

PR #1290, opened 2020-12-21
Footer update to align with design; Font update

leecalcote:
**Description**

This PR fixes #

**Notes for Reviewers**


leecalcote:
@Tanuj22 the qanelas font isn't (wasn't) being applied in many places. I wasn't sure how to import the app.style.js into the Footer component, and so I'm using straight .css. What do you think about this approach?

Nikhil-Ladha:
I think the app.style.js is automatically import d for all the pages. The reason for the correct font not being shown is because the correct tags might not be used for the same.

Like if you will see, currently we are showing the headings using just the link component which doesn't have the 'Qanelas Soft' font saved for it.

Tanuj22:
> I think the app.style.js is automatically import d for all the pages. The reason for the correct font not being shown is because the correct tags might not be used for the same.
> 
> Like if you will see, currently we are showing the headings using just the link component which doesn't have the 'Qanelas Soft' font saved for it.

Yeah. So if we want to use 'Qanelas Soft' for a text that isn't headings we will have to set the font-style for that in its respective css.

Nikhil-Ladha:
I think if we intend to use any particular font for any section we can use the appropriate tag which has that font associated with it.
Or assign font-family to the tags that should have a particular font-family globally, so as to keep a consistency across the site.

leecalcote (commented):
Merging progress... will follow-up with fixes if font is still off.

PR #1299, opened 2020-12-24
Updated SMI table for layer5-ng

dhruv0000:
Signed-off-by: dhruv0000 <patel.4@iitj.ac.in>

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/src/components/SMI-Table/LandscapeTable.style.js
+++ b/src/components/SMI-Table/LandscapeTable.style.js
@@ -1,0 +1,64 @@
+import styled from "styled-components";
+
+export const TableWrapper = styled.div`
+overflow-x: auto;
+
+table {
+    border-spacing: 0;
+    width: 100%;
+  
+    td {
+        padding: 5px;
+        height: 60px;
+        color: black;
+        text-align: center;
+        font-size: 14px;
+        font-weight: 400;
+        border-bottom: 1px solid #efefef;
+    }
+    th {
+        background: #1E2117;
+        color: #ffffff;
+        height: 60px;
+        font-weight: 600;
+    }
+  
+	tr.primaryRow {
+		background: #ffffff;
+		td {
+			border-right: 1px solid #ffffff;
+		}
+	}
+	tr.secondaryRow {
+		background: #efefef;
+		td {
+			border-right: 1px solid #efefef;
+		}
+	}
+
+	tr.secondaryRow-hidden {
+		display: none;
+	}
+
+    tr {
+		:last-child {
+			td {
+				border-bottom: 0;
+			}
+		}
+    }
+  
+    td, th {
+		:last-child {
+			border-right: 0;
+		}
+	}
+	
+	th, tr{
+		&:hover{
+			box-shadow: 0px 2px 15px -10px black;
+			transform: translateY(0px);
+		}
+	}
+  }
+`;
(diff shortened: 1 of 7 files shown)


src/components/SMI-Table/LandscapeTable.style.js
```diff
@@ -0,0 +1,64 @@
...
+export const TableWrapper = styled.div`
+overflow-x: auto;
+
+table {
+    border-spacing: 0;
+    width: 100%;
+  
+    td {
+        padding: 5px;
+        height: 60px;
```
Nikhil-Ladha, line 12: @dhruv0000 could you please update the CSS to use 'rem' unit instead of 'px'. Throughout the changes only you have done?

Nikhil-Ladha:
Also, I think the same CSS can be used for the smi table as well. I don't see any issues with that.

Nikhil-Ladha:
In Landscape-Grid/index.js at line 35, the 'tablist' for SMI Compatibility also needs to be removed.

Nikhil-Ladha (changes requested):
Please check the above comments.

src/components/SMI-Table/LandscapeTable.style.js
dhruv0000: Sure. But you might want to also look at the `Landscape-Table` CSS, on top of which I made some additional changes.

dhruv0000: Having said that, I don't think the modification I made on top of the Landsacpe-Table CSS had used 'px' anywhere.

dhruv0000:
> In Landscape-Grid/index.js at line 35, the 'tablist' for SMI Compatibility also needs to be removed.

yeah, sure. My bad.

src/components/SMI-Table/LandscapeTable.style.js
Nikhil-Ladha: Ohh, I missed that in my PR. If you don't mind could you please update the landscape table css as well to use 'rem' instead of 'px' ?

Also, then the css for both the tables would be same and the same file can be used for both instead of creating another one.

src/components/SMI-Table/index.js
```diff
@@ -4,7 +4,7 @@ import { FaCheckCircle } from "react-icons/fa";
 import { RiCloseCircleFill } from "react-icons/ri";
 import { IoMdHelpCircle } from "react-icons/io";
 import { IconContext } from "react-icons";
-import { TableWrapper } from "./LandscapeTable.style";
+import { TableWrapper } from "./SMITable.style";
```
Nikhil-Ladha, line 7: Can't the same stylesheet be used here as well?

dhruv0000: nope. Have made some updates in the SMI one. Having alternative color in SMI table rows wont make sense since the no. of rows is not constant.

Nikhil-Ladha: Why?
Like for other tables also we don't have constant rows.

Nikhil-Ladha: Infact I don't get the reason for creating a seperate file for the table as well. Even that can be reused.

leecalcote: @dhruv0000 we're in need of a multi-dimensional table. Are you thinking of adding a dropdown to choose between SMI versions?

leecalcote (approved):

leecalcote:
Merging progress... Two follow-up issues opened for go-live: #1302 and #1303

PR #1301, opened 2020-12-24
Updated the books page design

Nikhil-Ladha:

**Description**
This PR updates the book's page design.

**Note**:
The design is not exactly similar to Figma. Creating that design doesn't seem feasible so did a bit of modification to it.


src/sections/Books-grid/index.js
```diff
@@ -41,28 +40,60 @@ const BooksGrid = ({hide_path, limit}) => {
...
-                    </div>
-                </Container>
+            <PageHeader title="Service mesh with the best" path={path}/>
+            <h3 className="sub-head">Learn to service mesh from the world’s authority </h3>
+            <div className="book_list">
+                <Slider {...settings}>
+                    {data.allMdx.nodes.map(({id, frontmatter, fields }) => (
+                        <div className="book_image" key={id}>
+                            <div className="blog-slider_img">
+                                <img src={frontmatter.thumbnail.publicURL} alt={frontmatter.title} />
```
Tanuj22, line 91: Can't the `Image` component so gatsby image helps with image optimization?

src/sections/Books-grid/index.js
```diff
@@ -41,28 +40,60 @@ const BooksGrid = ({hide_path, limit}) => {
     );
 
     let path = hide_path ? "" : "Books";
-    let no_of_items = limit ? limit : data.length;
+    let thumb_imgs = [];
+    {data.allMdx.nodes.map(({frontmatter}) => (
+        thumb_imgs.push(frontmatter.thumbnail.publicURL)
+    ));}
+
+    var settings = {
```
Tanuj22, line 48: `const` should be used

src/sections/Books-grid/index.js
```diff
@@ -41,28 +40,60 @@ const BooksGrid = ({hide_path, limit}) => {
     );
 
     let path = hide_path ? "" : "Books";
-    let no_of_items = limit ? limit : data.length;
```
Tanuj22, line 44: This prop was passed down from pages/learn/index.js. Can you remove that also?

Tanuj22 (commented):
Shouldn't we include arrows in desktop view at the sides?

Nikhil-Ladha:
> Shouldn't we include arrows in desktop view at the sides?

I don't think it will be needed. As the user can drag, click to scroll through the carousel.

src/sections/Books-grid/index.js
Nikhil-Ladha: It was not helping, rather causing a lot of trouble. So had to use it this way, but will check again if it works.

src/sections/Books-grid/index.js
Nikhil-Ladha: It is done purposefully, because the same variable is used in the 'responsive' array within that variable.

src/sections/Books-grid/index.js
Nikhil-Ladha: I checked, it was never passed 😆

leecalcote:
We'll need to get better book cover images - both higher resolution and aspect ratio.

leecalcote:
> > Shouldn't we include arrows in desktop view at the sides?
> 
> I don't think it will be needed. As the user can drag, click to scroll through the carousel.

The thing is that most users won't realize that they can click and drag. I see the hand when hovering, but as a user am afraid to try to click and grab (because most websites don't have this functionality) for fear of the click causing me to be redirected to the specific book.

With only four books available, we'll want to disable infinite scroll. Otherwise it has a negative affect of exacerbating the fact that there are only four, not ten.

Along with this change to infinite: false, we have to be careful that if people can only see one book and not the edges of others, they may think there is only one book. The breakpoints and number of "slides" shown will need to be fiddled with, so that people are always seeing more than one book (partial of other books) no matter their screen size.

Nikhil-Ladha:
> We'll need to get better book cover images - both higher resolution and aspect ratio.

Yeah!

Nikhil-Ladha:
> The thing is that most users won't realize that they can click and drag. I see the hand when hovering, but as a user am afraid to try to click and grab (because most websites don't have this functionality) for fear of the click causing me to be redirected to the specific book.
> 
> With only four books available, we'll want to disable infinite scroll. Otherwise it has a negative affect of exacerbating the fact that there are only four, not ten.
> 
> Along with this change to infinite: false, we have to be careful that if people can only see one book and not the edges of others, they may think there is only one book. The breakpoints and number of "slides" shown will need to be fiddled with, so that people are always seeing more than one book (partial of other books) no matter their screen size.

I understand your point, but from the UX perspective I have applied enough CSS so, that doesn't happen. Like the grab, grabbing, pointer one.

But, still I will add the arrows and keep them intact as well, so that user can have both the experience of clicking and scrolling,  selecting and dragging.

The thing with removing infinite is:
1. The books will not be centered any more.
2. If I try to center, all the books will not be shown.

Will make the updates, and see if something good comes out of it.

leecalcote:
Let's merge and play with it a bit. See if arrows or the dots with mini book covers are necessary. The current preview is looking pretty great.

PR #1324, opened 2020-12-31
Added hover effect for the service mesh strength comparison section

Nikhil-Ladha:

**Description**


Nikhil-Ladha:
Only merge if we intend to keep this one over the current, or ignore 🙂

leecalcote:
At the end of the transition, there is a slight bump to the left for the comparative text. Weird.

leecalcote:
That slight bump at the end doesn't happen in Safari, but does in Chrome.

leecalcote (approved):

leecalcote:
@dhruv0000, please have a look at the SMI table. I tossed in a couple updates. They will be visible on https://layer5ng.netlify.app/projects/landscape soon.

src/sections/Landscape-grid/index.js
```diff
@@ -42,7 +42,7 @@ const LandscapeGrid=() => {
                                 <div className="comparison-mesh">
                                     <div className="mesh-img">
                                         <img src={Istio_Img} />
-                                        <h5>Istio</h5>
+                                        <h2>Istio</h2>
```
Nikhil-Ladha, line 45: I used h5, as it went with the front-size mentioned in the brand guide.
As per the brand guide h2 should have a font-size of 2.125 rem.

leecalcote: Oh, that makes sense. The font weight was in need of being adjusted, but as I went to make that adjustment, I arbitrarily and inappropriately changed to h2. I wonder which heading is the best-fit size for these.

Nikhil-Ladha: Ohh, I think the h5 heading goes well. Since it's a caption/name. I will update.

PR #1326, opened 2021-01-01
Few general updates

Nikhil-Ladha:


Jashpatel1 (commented):
I think decreasing the width in brand page in making it much longer(4 colors in row > 3 colors in row). The figma design shows it less wider with respect to bigger dimension.
Views @leecalcote @Tanuj22

Nikhil-Ladha:
> I think decreasing the width in brand page is making it much longer(4 colors in row > 3 colors in row). The figma design shows it less wider with respect to bigger dimension.
> Views @leecalcote @Tanuj22

I didn't get it. Are you suggesting to increase the width?

Nikhil-Ladha:
I think it's in some ways good that the page is long. It depicts how good our brand guide is and hoe much detailed analysis of logos and colors we have for different projects, programs.

Jashpatel1:
> > I think decreasing the width in brand page is making it much longer(4 colors in row > 3 colors in row). The figma design shows it less wider with respect to bigger dimension.
> > Views @leecalcote @Tanuj22
> 
> I didn't get it. Are you suggesting to increase the width?

Yes😅

Tanuj22 (approved):
For my part, I don't think the width should be increased more as the place will look too out of place compared to other pages then. 

We should later add a component like the one they have on the side here https://www.hashicorp.com/brand for easy navigation and more clear separation of sections

Nikhil-Ladha:
> For my part, I don't think the width should be increased more as the place will look too out of place compared to other pages then.
> 
> We should later add a component like the one they have on the side here https://www.hashicorp.com/brand for easy navigation and more clear separation of sections

Yeah, that would look much better.

Nikhil-Ladha:
Merging it, so that we can move ahead for the time being.

PR #1373, opened 2021-01-18
[NG] Update CONTRIBUTING.md

leecalcote:
**Description**
Updates to instructions for contributing on Layer5NG

- [ ] No, I didn't. GitHub web doesn't allow me to do so.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,15 +16,15 @@ Layer5 community members are an integral part of what makes Layer5 and it's proj
 
 The service mesh landscape is powered by Gatsbyjs. In order to add/update the landscape, fork this repository, clone it, create a branch and navigate to the **\_data/categories** folder. Edit these yaml files (if appropriate):
 
-- [non-functional.yml](https://github.com/layer5io/layer5/tree/layer5-ng/src/sections/Landscape-grid/non-functional.yml) - overview of various service meshes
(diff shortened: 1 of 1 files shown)


Nikhil-Ladha:
It would require some changes ig.....I think let's keep this PR open and on hold for the time being as we are almost there to go live with the site...once done I will push the updates to this PR....does that sound good?

leecalcote:
@Nikhil-Ladha, in general, that makes sense. This update is against the `layer5-ng` branch, though, where we already have a separate CONTRIBUTING.md that is specific to layer5-ng.  This PR is an update to that.

Nikhil-Ladha:
Okay...then in that aspect there might needs to be few updates in the landscape section of this guides as the files are no more in `yml` format but in `js` format 😅....also I think we should add the view only or view/comment link of the site design spec as the new members not already part of the community will not be able to access the guides then.

leecalcote:
Thanks for this, @Nikhil-Ladha. I changed from .yml to .js and added a note about the data source for the SMI table.

I'm ok with requiring that would-be contributors first join the community in order to gain access to resources like this Google Doc. Having the link here is helpful for those already part of the community and the lack of access hopefully serves as an incentive for would-be members to join.

Nikhil-Ladha (approved):
Cool 😎

PR #1404, opened 2021-01-23
enhance katacoda lab pages

vineethvanga18:

**Description**

**Notes for Reviewers**


Tanuj22:
@vineethvanga18 the builds should be active now. Can you push any commit here now to trigger the deploy ?

Nikhil-Ladha:
Looks very nice 👌🏼....just had few doubts regarding the design:
1. Shouldn't the `maximize` icon relocated beneath the service mesh image? Currently, it looks a bit conjusted to me 🤔.
2. For the mobile view, the width of drop-down showing the details of the labs could be increased a bit ig. Also, would you confirm if there is no white space in the drop-down on the right side.
3. The design is good, but me as a user it didn't strike me that the two buttons on the side to switch between two categories are actually `buttons`, they seemed more like a section label to me. Does that look similar to you guys as well? ( @Tanuj22 @Jashpatel1 @leecalcote ) Or is it only me😅

Tanuj22:
Agree with all the points @Nikhil-Ladha pointed out. The design is feeling too congested currently. Would it be better if the labs are present at the individual places only and not on the same page also?

Tanuj22:
Also should we remove individual labs from `Container` as I think it should take more space. Like this https://www.envoyproxy.io/try/getting-started

leecalcote:
I tossed in a little formatting. Among other issues that still remain, one is that of extra padding/margin onclick of a course tab.

leecalcote:
Great feedback from @Nikhil-Ladha and @Tanuj22. I think issues 2 and 3 might be solved now (there is still extra padding/margin where it shouldn't be). What do you think?

leecalcote:
Ah, geez. This merge was accidental.

Well... I suppose we'll keep iterating from here...

PR #1481, opened 2021-02-04
[NG] Projects_Megamenu

vineethvanga18:

**Description**

**Notes for Reviewers**
If a project is to have other than "/projects/project-slug", it needs to have an attribute "link" in the front-matter.


Tanuj22:
I think it is better to hardcode these in the mega menu. As creating src/collections/projects/meshery/index.mdx is creating an URL `/projects/meshery` which is not something we want

vineethvanga18:
In future, do all the projects have other '/projects/....' . If this is the case it is better to not create pages for projects. If we have a mix of '/projects/...' and separate pages, it would be difficult to hardcode everytime

vineethvanga18:
Also for community menu, all the posts are external links in collections. What should be done?

leecalcote:
For now, in order to go-live, let's go with hardcoding both.

Tanuj22:
> In future, do all the projects have other than '/projects/....' . If this is the case it is better to not create pages for projects. If we have a mix of '/projects/...' and separate pages, it would be difficult to hardcode everytime

yes. Can you do this? . Because `projects/meshery` is supposed to redirect and we do not want a separate page for it

leecalcote:
⚠️  Assigning high priority label. This is a go-live blocker.

leecalcote:
@vineethvanga18 will you consider advancing this PR instead - https://github.com/layer5io/layer5/pull/1489? This PR is derived from yours here.

vineethvanga18:
Sure, @leecalcote

Jashpatel1:
This can be closed. @vineethvanga18

vineethvanga18:
Yes, @Jashpatel1. Doing it.

PR #1524, opened 2021-02-11
changes for go live 

Tanuj22:
**Description**

This PR fixes #

**Notes for Reviewers**
- Fix some urls 
- Newcomers map
- Comment out the search box in the blog 
- Remove unused packages
- Clean some code
- 


Tanuj22:
I wasn't able to find a good way for downloading the brand files without the site taking a performance hit. The best would be to add a drive link I think. Maybe that's why they do the same here https://www.hashicorp.com/brand

Nikhil-Ladha (approved):
Then, the only way which seems possible is by using pure JS.

leecalcote:
Hold your horses.... you're saying that you can't signal to a web browser to not preload any resources referenced by an anchor tag?

Nikhil-Ladha:
It seems like that...🤔

PR #1574, opened 2021-02-19
Fix z-index of stewards section

Nikhil-Ladha:
This PR fixes the z-index in the Stewards section of the home page

### DIFF
--- a/src/sections/Home/Cloud-Native/cloud-native.style.js
+++ b/src/sections/Home/Cloud-Native/cloud-native.style.js
@@ -11,17 +11,10 @@ const CloudNativeWrapper = styled.div`
      
     padding: 1rem 0rem 2rem 0rem;
     color:white;
-    margin-top:5rem;
-    margin-bottom: -9rem;
-    margin-left: 0rem;
-    margin-right: 0rem;
-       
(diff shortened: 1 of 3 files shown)


leecalcote:
Believe it not, I have the Cloud Native Leaders bleeding into Proudly Maintaining and Industry Stewards on purpose. Understanding that this is on purpose, wdyt about how this looks?

Nikhil-Ladha:
It is still bleeding into the stewards section. The issue is that when you try to hover on the `See out projects` button in the steward's section that is not working, because the z-index is lesser than 0, and the base/body has a higher z-index.

Nikhil-Ladha:
> Believe it not, I have the Cloud Native Leaders bleeding into Proudly Maintaining and Industry Stewards on purpose. Understanding that this is on purpose, wdyt about how this looks?

Currently, it doesn't bleed into the proud maintainers section

Nikhil-Ladha:
Can we get this merged? Considering the issue being on the homepage itself.

Tanuj22 (approved):

leecalcote:
Gents, thank you for moving forward here. 👍

PR #1644, opened 2021-03-01
CI: Site preview

Nikhil-Ladha:

### DIFF
--- a/.github/workflows/build-and-preview-site.yml
+++ b/.github/workflows/build-and-preview-site.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Deploy and Preview 🚀
         uses: JamesIves/github-pages-deploy-action@4.0.0
         with:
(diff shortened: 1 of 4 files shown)


Nikhil-Ladha:
@Tanuj22 any idea why the second workflow is not starting after one of the workflow finishes?

Tanuj22:
> @Tanuj22 any idea why the second workflow is not starting after one of the workflow finishes?

It needs to be merged first? Should I merge this?

Nikhil-Ladha:
Oh, so if we merge and then I modify both the workflows then it should work, right?

Nikhil-Ladha:
Or, again it will work after a merge?

Tanuj22:
Not sure

Nikhil-Ladha:
> Not sure

Okay, let's merge then we can check that..

Tanuj22 (approved):

PR #1948, opened 2021-06-25
Smooth transition for installation steps component

### LINKED ISSUE #1752: Smooth Transition of "Choose Your Platform" Command div
**Current Behavior**

Command div neither smoothly opens nor does it close.

**Desired Behavior**

Command div should open & close smoothly.

---
[Optional] **Screenshots / Mockups**

https://user-images.githubusercontent.com/44670961/113507184-9b525880-9566-11eb-839e-0b44334f174b.mp4

[Optional] **Alternatives**

---
**Contributor Resources**

- See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)
- See Layer5 site designs in this [Figma project](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs). Join the [Layer5 Community](http://slack.layer5.io) for access.

sudo-NithishKarthik:
Signed-off-by: suod-NithishKarthik <nithishkarthik01@gmail.com>

fix[site] : Add smooth transition for platform installation component

**Description**
Change the height of the component based on the state of platform
availability and add transition property to the component

This PR fixes #1752

**Notes for Reviewers**
The height has been set to `200px` which i am not totally OK with. I am afraid that in some cases, the the height of the content might become more than that and because `overflow` is set to `hidden`, it might not be fully visible. 

### DIFF
--- a/src/sections/Meshery/Meshery-platforms/index.js
+++ b/src/sections/Meshery/Meshery-platforms/index.js
@@ -174,13 +174,27 @@ const supported_platforms = [
 
 const MesheryPlatforms = () => {
   const [currentPlatform, setCurrentPlatform] = useState({});
+  const [installationStepsHeight,setInstallationStepsHeight] = useState(currentPlatform.name ? "200px" : 0);
 
-  const changeCurrentPlatform = (index) => {
-    if (currentPlatform.name && currentPlatform.name === supported_platforms[index].name)
+  const hasSelectedSamePlatform = (index) => currentPlatform.name === supported_platforms[index].name;
+
+  const changeCurrentPlatformState = (index) => {
+    if (currentPlatform.name && hasSelectedSamePlatform(index))
       setCurrentPlatform({});
     else
       setCurrentPlatform(supported_platforms[index]);
+
   };
+
+  const changeCurrentPlatform = (index) => {
+    if(currentPlatform.name && !hasSelectedSamePlatform(index)) {
+      changeCurrentPlatformState(index);
+      return;
+    }
+    setTimeout(() => changeCurrentPlatformState(index), 500);
+    setInstallationStepsHeight(currentPlatform.name ? 0 : "200px");
+  };
+
 
   return (
     <MesheryPlatformsWrapper>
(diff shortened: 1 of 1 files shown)


l5io:
🚀 Preview for commit 23c90a01d9964b17759e22d36439fe6ca203893d at: https://60d562c09790914f3c8be569--layer5ng.netlify.app

Nikhil-Ladha:
You can use `min-height` instead, that would work for all such cases.

sudo-NithishKarthik:
> You can use `min-height` instead, that would work for all such cases.

Actually, that will introduce a delay on one side, it wont be smooth 😅 , that's why.
Ill try to implement it using JS since we don't want to get bothered by this after sometime 👍🏽

l5io:
🚀 Preview for commit 23c90a01d9964b17759e22d36439fe6ca203893d at: https://60d5eec331beeb5936ce0c01--layer5ng.netlify.app

Nikhil-Ladha:
@sudo-NithishKarthik is the PR complete are you still experimenting with dynamic values?
The reason I am fine with the current fixed height is that it works for all the platforms that we have now. And, new platforms is not something that happens always. We can live with this code for sure.

sudo-NithishKarthik:
Ok, in that case, we should be good to go 👍🏽

src/sections/Meshery/Meshery-platforms/index.js
```diff
@@ -174,14 +174,28 @@ const supported_platforms = [
...
       setCurrentPlatform(supported_platforms[index]);
+
   };
 
+  const changeCurrentPlatform = (index) => {
+    if(currentPlatform.name && !hasSelectedSamePlatform(index)) {
+      changeCurrentPlatformState(index);
+      return;
+    }
+    setTimeout(() => changeCurrentPlatformState(index), 500);
```
Nikhil-Ladha, line 194: Why do we need this setTimeout though? We have already added a delay in the CSS below.

sudo-NithishKarthik: That's because, without that timeout, the container would be unmounted instantly, so the transition will not take place

Nikhil-Ladha: Instead of timeout use `useRef` to add a class for expansion and remove it otherwise. `setTimeout` is not the best choice here. Then, you might even need animation for the same or if the transition property works, then that's fine as well.

sudo-NithishKarthik: Actually, the issue is that, we are conditionally rendering the container right, so, without `setTimeout`, the desired outcome cannot be achieved ig. For instance, lets say i the user wants to close the container, he will click the button, now, when we use `ref` to add/remove class to it, then, we can't get the animation to work because, the moment he presses the button, the container will be unmounted and the `ref.current` will point to `undefined`

Nikhil-Ladha: Ack.... let's keep the setTimeout for the time being. Since, this is restricted to a very small portion of the site it is not much of a problem.

Nikhil-Ladha (approved):

PR #2220, opened 2021-09-25
Added changes in contributing.md regarding the make command and fixed…

prashant-isbest:
… the naming of the checkout branch from which pull request will be send

Signed-off-by: Prashant Joshi <scienceprashant27@gmail.com>

**Description**

This PR fixes # contributing.md

**Notes for Reviewers**

- [ yes] Yes, I signed my commits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,17 +80,18 @@ Please contribute! The Layer5 site uses Gatsby. The process of contributing to d
 1. Get a local clone of your fork of the site.
 `git clone https://github.com/layer5io/layer5`
 1. Switch to the `master` branch.
-`git checkout master`
+`git switch master`
 1. Create and checkout a new branch to make changes within
-`git checkout -b <my-username>/<my-changes>`
+`git checkout -b <my-changes-branchname>`
 1. Edit site/add content.
 `vi <specific page>.md` # or use your favorite IDE
 1. Run site locally to preview changes.
 `make site` # this will run a local web server with "live reload" conveniently enabled.
+(NOTE make is a linux command , if you are using windows you can install wsl , and then run command `sudo apt install make` )
 1. Commit your changes to your remote branch.
 `git commit --signoff -m"<commit subject>`
 1. Push your changes
-`git push origin <my-username>/<my-changes>`
+`git push origin <my-changes-branchname>`
 1. Open a [pull request](../../pulls) (in your web browser) against the `master` branch on https://github.com/layer5io/layer5.
 
 ### <a name="commit-signing">Signing-off on Commits (Developer Certificate of Origin)</a>


l5io:
🚀 Preview for commit 07fb442d37271dc630cbbe9c2f86f59a400d7b06 at: https://614f511fa494f69311915bf3--layer5ng.netlify.app

debo19 (approved):
LGTM 🎉

warunicorn19 (approved):
Thank you @prashant-isbest.😄

CONTRIBUTING.md
```diff
@@ -80,17 +80,18 @@ Please contribute! The Layer5 site uses Gatsby. The process of contributing to d
 1. Get a local clone of your fork of the site.
 `git clone https://github.com/layer5io/layer5`
 1. Switch to the `master` branch.
-`git checkout master`
+`git switch master`
```
Nikhil-Ladha, line 83: Why this change?

CONTRIBUTING.md
```diff
@@ -80,17 +80,18 @@ Please contribute! The Layer5 site uses Gatsby. The process of contributing to d
...
 1. Create and checkout a new branch to make changes within
-`git checkout -b <my-username>/<my-changes>`
+`git checkout -b <my-changes-branchname>`
 1. Edit site/add content.
 `vi <specific page>.md` # or use your favorite IDE
 1. Run site locally to preview changes.
 `make site` # this will run a local web server with "live reload" conveniently enabled.
+(NOTE make is a linux command , if you are using windows you can install wsl , and then run command `sudo apt install make` )
```
Nikhil-Ladha, line 90: I doubt that!
`make` do work in windows as well.
 We have GNU make for Windows.

CONTRIBUTING.md
warunicorn19: @Nikhil-Ladha actually if we don't do `git switch master` the previous commits of a different branch also get included into the new one.

Nikhil-Ladha: Which branch and what commits? Example?

warunicorn19: For example, if I'm already working on this repo and I create a branch to work on some issue, I do the changes and make a PR, meanwhile I have to work on another issue and so I create another branch to work on, before creating a new branch if I don't do `git switch master` all my changes from the previous branch would also come onto the new branch.

warunicorn19: Though ig we would probably not need this information here considering this is for getting started, but then we could also get rid of the `git checkout master` here cause after cloning we would be on the master branch itself 🤔, right? @Nikhil-Ladha

Nikhil-Ladha: That's obvious.
But, the same can be done with `git checkout master`

warunicorn19: Oh, I didn't know that😅, I've actually always done `git switch master` just to be safe.

Nikhil-Ladha: Ok, so the thing is `git checkout master` does exactly everything of `git switch master` in fact, git` switch` is a (code)copy of `git checkout master`. `git switch` was added considering the dual nature of `git checkout` to create branches and restore changes. And, therefore to avoid confusions 2 new commands `git switch` and `git restore` was added.

Nikhil-Ladha: > Oh, I didn't know thatsweat_smile, I've actually always done git switch master just to be safe.
It's fine. Just, that the such PRs should have a better commit-msg and description to justify the changes being done.

warunicorn19: Aah okay, makes sense, yes you're right.

PR #2409, opened 2021-12-10
[Cl] Improve Newcomers Alert Workflow

theBeginner86:
Signed-off-by: Pranav Singh <pranavsingh02@hotmail.com>

**Description**
- In this PR, I've improved the working of the newcomers-alert workflow. 

*Why workflow was commenting multiple times on Slack?*
- The workflow is triggered on the following scenarios: opening a new issue or adding labels to existing issues (allowed labels: `good first issue` or `first-timers-only`). 
- Initially, I wasn't comparing the labels with the right metadata (I was comparing with an array that contains data for all the labels added to the issue which triggered the workflow) but since that array was the updated one, with our specified label already there so the each of the triggered passes the workflow checks and the message was being printed multiple times.
- Now, instead of checking with the array, I'm checking with the label name of the event itself which triggered the workflow. So, only the one which contains our specified label would pass the if condition and print the message on slack

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/newcomers-alert.yml
+++ b/.github/workflows/newcomers-alert.yml
@@ -4,7 +4,7 @@ on:
     types: [opened, labeled]
 jobs:
   good-first-issue-notify:
-    if: contains(github.event.issue.labels.*.name, 'good first issue') || contains(github.event.issue.labels.*.name, 'first-timers-only') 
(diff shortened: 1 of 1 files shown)


warunicorn19 (approved):
Looks great! The description really helps. Thank you @theBeginner86

PR #2539, opened 2022-01-25
Build: Added 'cache-preserve' flags to optimise build time

Nikhil-Ladha:

**Description**
Added flags in the `gatsby.config` file to preserve the cache generated from the previous build to optimize
the time taken during a deploy, or development preview.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -9,6 +9,11 @@ module.exports = {
     image: "/images/layer5-company.png",
     twitterUsername: "@layer5",
   },
+  flags: {
+    FAST_DEV: true,
+    PRESERVE_WEBPACK_CACHE: true,
+    PRESERVE_FILE_DOWNLOAD_CACHE: true,
+  },
   plugins: [
     "gatsby-plugin-sitemap",
     "gatsby-plugin-react-helmet",


Nikhil-Ladha:
Locally I could see a difference of ~60secs for prod build

l5io:
🚀 Preview for commit 84aa66a1d3e1ad9f4a9a9819fcb5bd0229c26a74 at: https://61ef9c63b3854e9d89007121--layer5ng.netlify.app

leecalcote:
Good idea. I wonder what the behaviors are around these caches. I wonder when/if we'll run into weirdness caused by a cache not being busted when it should have been. This wondering doesn't mean that we shouldn't use the caching feature, but that it would be good to understand potential caveats to look out for.

leecalcote (approved):

Nikhil-Ladha:
> Good idea. I wonder what the behaviors are around these caches. I wonder when/if we'll run into weirdness caused by a cache not being busted when it should have been. This wondering doesn't mean that we shouldn't use the caching feature, but that it would be good to understand potential caveats to look out for.

Agreed. It actually works keeping that use case in mind, and updates the cache if any data/file/asset has been updated else just skips the re-build of the files.

Nikhil-Ladha:
Merging the PR. If (hopefully we shouldn't) we face any corner issues in future, we can disable/update as per need.

PR #2566, opened 2022-02-01
Addition of robots.txt

### LINKED ISSUE #2565: Missing robots.txt
It's hard to believe that we've been missing this critical file. A prompt fix is needed here with some follow-up study of best practices for this file's structure.

leecalcote:
**Description**

This PR fixes #2565 

**Notes for Reviewers**
We might want to define DEV and PROD `env` in this plugin's config.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -445,6 +445,15 @@ module.exports = {
         query: "allMdx",
       },
     },
+    {
+      resolve: "gatsby-plugin-robots-txt",
+      options: {
+        host: "https://layer5.io",
+        sitemap: "https://layer5.io/sitemap.xml",
+        policy: [{userAgent: "*", allow: "/"}],
+      }
+    },
     "gatsby-plugin-meta-redirect", // make sure this is always the last one
   ],
+  
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "gatsby-plugin-preload-fonts": "^2.14.0",
         "gatsby-plugin-react-helmet": "^4.14.0",
         "gatsby-plugin-react-helmet-canonical-urls": "^1.4.0",
+        "gatsby-plugin-robots-txt": "^1.7.0",
         "gatsby-plugin-sharp": "^3.14.3",
         "gatsby-plugin-sitemap": "^3.3.0",
         "gatsby-plugin-styled-components": "^4.14.0",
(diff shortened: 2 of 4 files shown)


Nikhil-Ladha:
We are already adding robots.txt for netlify builds. What is the purpose of this plugin?

leecalcote:
@Nikhil-Ladha, try accessing https://layer5.io/robots.txt

Nikhil-Ladha:
Yeah, that is not there....but robots.txt file is generally used for exclusion is what I know, that's the reason I don't understand the reason of adding it.
Is it to only crawl the sitemap of the site?

l5io:
🚀 Preview for commit 5ad5af010d137707a49873a69aedb8bc9cca4548 at: https://61f8a82f385f9620468e81d1--layer5ng.netlify.app

leecalcote:
@Nikhil-Ladha, you're right. There are a few reasons why having one defined is better than not having one defined. None of which, in so much as I understand, are strong reasons:

As a future-facing concern, we will want to _maximize crawl budget_. Context: "If you’re having a tough time getting all of your pages indexed, you might have a crawl budget problem. By blocking unimportant pages with robots.txt, Googlebot can spend more of your crawl budget on the pages that actually matter."

We have a copy of one of our ebooks hanging out in a public directory as a PDF, so we might want to _prevent indexing of resources_: "Using meta directives can work just as well as Robots.txt for preventing pages from getting indexed. However, meta directives don’t work well for multimedia resources, like PDFs and images. That’s where robots.txt comes into play."

Prevent SEO tools from notifying that we're missing this file.

leecalcote:
I can't tell if the plugin is working as desired, because it didn't generate a robots.txt file locally, and the disallow robots.txt that we are generating in deploy previews might be overwriting one potentially generated by the plugin.

Nikhil-Ladha:
Yeah, that might be the case for deploy previews as we are explicitly creating a robots file there.
Though, I just went through the plugin docs and I see it provides additional options to set this for netlify previews, so maybe we can explore that possibility and use it to the best extent and avoid this collision.

leecalcote:
@Nikhil-Ladha to your point, we could forego using a plugin and simply statically define and publish one. It's unclear to me that missing one, currently, is doing any damage.

Nikhil-Ladha:
Let's merge this, then I can send out a following PR which would incorporate all the above points.
That is, prod/dev config and the netlify previews config as well.

Nikhil-Ladha (approved):

leecalcote:
> Yeah, that might be the case for deploy previews as we are explicitly creating a robots file there.
> Though, I just went through the plugin docs and I see it provides additional options to set this for netlify previews, so maybe we can explore that possibility and use it to the best extent and avoid this collision.

Right on. 👍 If it does handle ENVs / Netlify deploy preview blocking (consolidating these concerns into one plugin config) - that might be nice.

Also, if not using the plugin feels better, that'll work as well. I don't have much of an opinion here.

Nikhil-Ladha:
> > Yeah, that might be the case for deploy previews as we are explicitly creating a robots file there.
> > Though, I just went through the plugin docs and I see it provides additional options to set this for netlify previews, so maybe we can explore that possibility and use it to the best extent and avoid this collision.
> 
> Right on. 👍 If it does handle ENVs / Netlify deploy preview blocking (consolidating these concerns into one plugin config) - that might be nice.
>

Sure.

> Also, if not using the plugin feels better, that'll work as well. I don't have much of an opinion here.

Ahh....no worries. Let's not allow _any_ report out there to show any such _missing_ error 😄

leecalcote:
I have to admit that the above reasons aren't terribly compelling. It is nice to put a tick in the missing file box, though. :)

PR #2874, opened 2022-06-07
[UI]Added Cookies for banner

Harshit-Dandriyal:
Signed-off-by: Harshit Dandriyal <56961488+Harshit-Dandriyal@users.noreply.github.com>

**Description**
Created cookies for previous Banner
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,12 +24,15 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@emotion/react": "^11.9.0",
+    "@emotion/styled": "^11.8.1",
     "@fullcalendar/daygrid": "^5.10.1",
     "@fullcalendar/google-calendar": "^5.10.1",
     "@fullcalendar/interaction": "^5.11.0",
     "@fullcalendar/react": "^5.10.1",
     "@mdx-js/mdx": "^1.6.22",
     "@mdx-js/react": "^1.6.22",
+    "@mui/material": "^5.8.2",
     "@react-icons/all-files": "^4.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,12 +24,15 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@emotion/react": "^11.9.0",
+    "@emotion/styled": "^11.8.1",
```
Nikhil-Ladha, line 28: Why do we need these 2 packages and the one below `@mui/material` for?

l5io:
🚀 Preview for commit 19de4f1c3cae0abe458b6e2e7f00929f60b4f57c at: https://629fa7114d130d0e1bda059e--layer5.netlify.app

package.json
Harshit-Dandriyal: Actually, I was working on a dark theme, yes but don't need it in this PR.

Harshit-Dandriyal:
@Nikhil-Ladha Here our cookies are being retrieved after loading the components, It is not working as expected.

Nikhil-Ladha:
Here the first load takes some good time, later it is faster.
Also, one major drawback is that it is not rendering at all, though again there can be some initial render but again it would be the same as we have now.

Nikhil-Ladha:
I think we can let it be as it is now, until and unless anyone manually does a couple of reloads on the home page they would never see that issue, and highly doubt that situation will ever occur (Maybe 1 in a 100k or so?)

PR #3013, opened 2022-08-03
Updated the sitemap path to root directory

Nikhil-Ladha:

**Description**
Updated the sitemap path to `/sitemap-index.xml`.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -21,7 +21,12 @@ module.exports = {
         disable: true
       }
     },
-    "gatsby-plugin-sitemap",
+    {
+      resolve: "gatsby-plugin-sitemap",
+      options: {
+        output: "/",
+      }
+    },
     "gatsby-plugin-react-helmet",
     {
       resolve: "gatsby-plugin-react-helmet-canonical-urls",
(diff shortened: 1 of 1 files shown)


Nikhil-Ladha:
@leecalcote we need to update the Google Search Console as well, if we need this change.

l5io:
🚀 Preview for commit 86c87632241503ba01fe6d410cb73676be4e121b at: https://62ea3637f354f07ae5d1504e--layer5.netlify.app

leecalcote:
This is good. Maybe a short chat on this on the next Websites call to clarify our permanent configuration.

l5io:
🚀 Preview for commit faa62220b7a81e825ba9b051e009810dce3b87c1 at: https://62eaadfac810e80944c809ee--layer5.netlify.app

leecalcote:
I was going to be picky about cluttering the root directory with an index file and then another sitemap-0 file instead of having them both in a sitemap folder. Ultimately, whether we stay as-is or use this configuration, it doesn't really matter. I'll let you make the call. If we do go forward here, you're right, we'll need to update this:
Screen Shot 2022-08-03 at 12 59 46 PM

Nikhil-Ladha:
> I was going to be picky about cluttering the root directory with an index file and then another sitemap-0 file instead of having them both in a sitemap folder. Ultimately, whether we stay as-is or use the configuration in this PR doesn't really matter. I'll let you make the call. If we do go forward with the config change in this PR, you're right, we'll need to update this:

Agree with your point. TBH, I don't really mind the current configuration of `sitemap/sitemap-index.xml` it is kind of cleaner. I think we can let it be as it is, and avoid too many change to sitemap URLs :)

debo19:
If we cannot have `/sitemap.xml` back, I think keeping both the files under `sitemap` folder is good enough since it doesn't hurt the SEO.

debo19:
Yikes I'm late 😆

Issue #2943, opened 2022-06-28
[Feature] Integrations Collection


leecalcote:
**Current Behavior**
Meshery has any number of integrations with other technologies and that number will only grow. While we list the 10 or so service meshes that Meshery integrates with, however, there will be many, many others, like Kubernetes, SPIRE, Argo, Prometheus, Grafana, Jaeger, CloudEvents, and so on…

--> _Meshery needs a catalog of integrations for users to browse._

**Desired Behavior**
1. an `integrations` collection.
  1. integrations should be categorized.
1. a new `/cloud-native-management/meshery/integrations` page with integrations overview and having a sortable grid (or similar) perusable catalog of integrations. 
    1. Ideally, this is something less boring that a blocky grid of cards.
1. an individual page for each integration, describing the functionality of the integration.

**Inspiration**
Design inspiration can be taken from relevant solutions like:

1. https://www.bigpanda.io/integrations
1. https://www.datadoghq.com/product/platform/integrations/#all
1. https://slack.com/integrations
1. https://www.dynatrace.com/hub/
1. A honeycomb-style example:
![image](https://user-images.githubusercontent.com/7570704/176253337-06a57cd5-e8c2-48b5-9ecc-8d9f227244c3.png)

---

<img src="https://raw.githubusercontent.com/layer5io/layer5/master/.github/assets/images/layer5/5-light-small.svg" width="16px" align="left" /><h4>Contributor Resources</h4>

The layer5.io website uses Gatsby, React, and GitHub Pages. Site content is found under the [`master` branch](https://github.com/layer5io/layer5/tree/master).
- See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)
- See Layer5 site designs in this [Figma project](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs). Fill-in the [Layer5 Community Member Form](https://layer5.io/newcomer) and join the [Layer5 Community Slack](http://slack.layer5.io) for access.

sahilsuman933:
@leecalcote could you please assign this issue to me. I really wanted to work on this issue.

leecalcote:
Sounds good, @sahilsuman933 👍

PR #3059, opened 2022-08-10
[Meshery] Fixed Individual Integration Page Style

sahilsuman933:
**Description**

- Fixed Individual Integration Page responsive style
- Added CTA on Integration Collection Page

This PR updates #2943 

**Notes for Reviewers**


Nikhil-Ladha:
While you create another PR, make sure never to use "spaces" in file/folder names. Use "-" (hyphens) instead. Like, "App-Mesh"

Nikhil-Ladha:
Also, no need to add 3 slides for all the integrations until we have the real one's. It is unnecessary, just add one for reference. We can keep 3 in docker to show how the slider is actually

PR #3091, opened 2022-08-18
An attempt to adjust webp image quality for layer5-projects.svg

leecalcote:
Help needed.

Signed-off-by: Lee Calcote <leecalcote@gmail.com>

**Description**

This doesn't work, yet...

**Notes for Reviewers**

Help needed.

### DIFF
--- a/src/sections/Company/About/index.js
+++ b/src/sections/Company/About/index.js
@@ -14,9 +14,37 @@ import CNCFstackedlogo from "./images/cncf-stacked-color.svg";
 import CNCFLightstackedlogo from "./images/cncf-light-stacked-color.svg";
 
 import { Link } from "gatsby";
+import { graphql, useStaticQuery } from "gatsby";
+import { getImage } from "gatsby-plugin-image";
+import { BgImage } from "gbimage-bridge";
+
 const Community_meetup = "./images/Community_meetup.png";
-const Layer5Projects = "./images/layer5-projects.png";
+// const Layer5Projects = "./images/layer5-projects.svg";
+
+
+
 const About = ({ theme }) => {
+
+  // Lee Calcote learned this trick from Maintainer Nikhil Ladha
+  const { webpImage } = useStaticQuery(
+    graphql`
+      query {
+        webpImage: file(
+          relativePath: { eq: "./images/layer5-projects.svg"" }
+        ) {
+          childImageSharp {
+            gatsbyImageData(
+              width: 2000
+              quality: 50
+              webpOptions: { quality: 70 }
+            )
+          }
+        }
+      }
+    `
+  );
+
+  const Layer5Projects = getImage(webpImage);
   return (
     <AboutSectionWrapper id="about">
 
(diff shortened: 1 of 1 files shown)


src/sections/Company/About/index.js
```diff
@@ -14,9 +14,37 @@ import CNCFstackedlogo from "./images/cncf-stacked-color.svg";
...
+
 const Community_meetup = "./images/Community_meetup.png";
-const Layer5Projects = "./images/layer5-projects.png";
+// const Layer5Projects = "./images/layer5-projects.svg";
+
+
+
 const About = ({ theme }) => {
+
+  // Lee Calcote learned this trick from Maintainer Nikhil Ladha
```
Nikhil-Ladha, line 28: Ha! 😂

src/sections/Company/About/index.js
```diff
@@ -14,9 +14,37 @@ import CNCFstackedlogo from "./images/cncf-stacked-color.svg";
...
+  const { webpImage } = useStaticQuery(
+    graphql`
+      query {
```
sahilsuman933, line 31: You can use this query to fetch all the images after adding the path of about images in the gatsby-config.js file. 

`  {
      resolve: "gatsby-source-filesystem",
      options: {
        name: "about-images",
        path: `${__dirname}/src/sections/Company/About/images`,
      },
    },`



`{
  allFile(filter: { sourceInstanceName: { eq: "about-images" } }) {
    nodes {
      name
      childImageSharp {
        gatsbyImageData(width: 2000, quality: 50, webpOptions: { quality: 70 })
      }
    }
  }
}
`

src/sections/Company/About/index.js
```diff
@@ -14,9 +14,37 @@ import CNCFstackedlogo from "./images/cncf-stacked-color.svg";
...
+        webpImage: file(
+          relativePath: { eq: "./images/layer5-projects.svg"" }
```
Nikhil-Ladha, line 33: If we are using the gatsby-plugin, then let's use the PNG image itself.
SVG doesn't work correctly with the plugin.

Nikhil-Ladha: And, since it is a relative path, we don't need to provide the folder name if it's under the root (here `images`) directory.
i.e, we can simply use `layer5-projects.png`

src/sections/Company/About/index.js
```diff
@@ -41,7 +69,8 @@ const About = ({ theme }) => {
                 </div>
               </Col>
             </Row>
-            <StaticImage src={Layer5Projects} alt="About Layer5 Projects" />
+            <BgImage image={Layer5Projects} className="section" />
```
Nikhil-Ladha, line 72: We are missing the closing tag for `BgImage` 😬

src/sections/Company/About/index.js
Nikhil-Ladha: Naa, not need to use it that way when we are querying using the relative path key.
It can simply be a file name, where the path should be the relative path from the root directory of the sourced directories.
In this case simply `layer5-projects.png`

src/sections/Company/About/index.js
sahilsuman933: Also, we have a extra double quote at the end of the relativePath

Nikhil-Ladha (commented):
I would say if we have the SVG version of the image, then let's skip the plugin usage completely and simply render the SVG image using the traditional `img` element. It's totally fine and no harm to the page load.

leecalcote:
Thanks for entertaining this PR. We're good to close out this experiment, I think.

Issue #3154, opened 2022-09-06
[Blog] Make category and tag list scrollable


debo19:
**Current Behavior**
With the ever-growing tags and category list for blogs, the sidebar is becoming too long and is overlapping with the footer.

<img width="596" alt="blog-tag-issue" src="https://user-images.githubusercontent.com/85789734/188614144-ffc86877-79b2-43bc-bb93-2aaf14bab43d.png">

**Desired Situation**
Give the tag and category list a fixed height on the page but make both the the lists scrollable.

---
**Contributor Resources**

The layer5.io website uses Gatsby, React, and GitHub Pages. Site content is found under the [`master` branch](https://github.com/layer5io/layer5/tree/master).
- See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)
- See Layer5 site designs in this [Figma project](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs). Join the [Layer5 Community](http://slack.layer5.io) for access.

NPDebs:
Hi. 
Can I work on this task?

debo19:
Sure @NPDebs

NPDebs:
Thank you! 🤲🏻🙂😊

zwwang98:
Hello! May I join you @NPDebs to work on this issue together?

vishnus17:
Hi @NPDebs. 
Are you still working on this issue? If not, can I take over ? I was going through it and I've found a solution. 
Thanks.

debo19:
@zwwang98 Since you commented first, I'll wait for you to respond if you want to take this up. Otherwise we'll assign @vishnus17

zwwang98:
> @zwwang98 Since you commented first, I'll wait for you to respond if you want to take this up. Otherwise we'll assign @vishnus17

I would love to work on it. Really appreciate for this opportunity.

debo19:
Go ahead @zwwang98

zwwang98:
> Go ahead @zwwang98

Hello @debo19 ! Thank you again for this opportunity! I really appreciate it!  

I made some progress and I want to make sure I am on the right track:
1. added another class named `scroll` for the list `ul`
2. set the class with `max-height` and `overflow` in the `style.js` file

Here is how it looks now.
[...]

PR #3169, opened 2022-09-11
fix sidebar and footer collapsing issue #3154

vishnus17:
**Description**

This PR is a solution for #3154

**Notes for Reviewers**


vishnus17:
image

This is a short solution. Added bottom padding to prevent footer collapsing on subscribe component.

Nikhil-Ladha:
This is not the intended solution for this issue.
We need to make the `tags` list have a `max-height` height, with `overflow: auto`, so that in case of overflow we have a scrollbar.

Though, I don't think the issue is assigned to you. So, please close the PR or confirm with the assignee of the issue regarding your interest, before sending a PR on an assigned issue. In case, of an un-assigned issue get the issue assigned to yourself by commenting on it.

vishnus17:
Okay.

PR #3344, opened 2022-10-17
Fix the build / Preview issue

sahilsuman933:
**Description**

**Notes for Reviewers**


sahilsuman933:
@Nikhil-Ladha I fixed the build issue that was caused due to the Integration page, but we are now getting another issue due to the plugin. So, could you please have a look at it?

sahilsuman933:
Yeah, I added `?` to ignore the execution and return null/undefined to prevent the failure of making the build via GitHub action. Also, I did check the local build, and it worked. 

So, in your opinion, should I use the "?" or not? @Nikhil-Ladha

Nikhil-Ladha:
Nope, you shouldn't.
Always try to find the actual root of the issue :)
Here the issue is with the path of the `index.mdx` file of crossplane. It is mistakenly added under the `icons` folder, rather it should be under the `crossplane` root folder under the integrations collection.
Ref: https://github.com/layer5io/layer5/tree/master/src/collections/integrations/crossplane/icon

Fixing the path of index.mdx will resolve the issue.

l5io:
🚀 Preview for commit 09338cf18fc200c774b25b95ee22f0dea85c62c1 at: https://634d59fe418531322e53d5d6--layer5.netlify.app

Issue #3338, opened 2022-10-16
Image is not aligned with the text on page: /projects/nighthawk


Krishna2323:
#### Description
Image is not aligned with the text on page: /projects/nighthawk
![Screenshot 2022-10-16 222936](https://user-images.githubusercontent.com/85894871/196048267-4117207b-0e23-40dd-a11d-e73ce15741e7.png)

#### Expected Behavior
![Screenshot 2022-10-16 222826](https://user-images.githubusercontent.com/85894871/196048273-4857d954-19ce-4a67-9b48-089d5aeb2d98.png)

#### Environment:
- Host OS: 
- Browser: 

---
<img src="https://raw.githubusercontent.com/layer5io/layer5/master/.github/assets/images/layer5/5-light-small.svg" width="16px" align="left" /><h3> Contributor Resources and <a href="https://layer5.io/community/handbook">Handbook</a></h3>

The layer5.io website uses Gatsby, React, and GitHub Pages. Site content is found under the [`master` branch](https://github.com/layer5io/layer5/tree/master).
- 📚 See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md)
- 🎨 Wireframes and designs for Layer5 site in [Figma](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs). 
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

Nimesh-21:
Can you please assign this to me? I am new to open source.

mritul:
@Krishna2323  Can i take up this issue and fix it ?

NnajiChimuanya:
Assign me this @Krishna2323

AbhishekTiwari23:
My setup is also ready.

legendarykamal:
@Nimesh-21 @NnajiChimuanya please understand that in order to be assigned, 
you must have atleast your setup ready (a primary requirement) other requirements could be a basic idea of how gatsby works.

I wonder if you have your setup done. ? 🤔

Nimesh-21:
> @Nimesh-21 @NnajiChimuanya please understand that in order to be assigned, 
> you must have atleast your setup ready (a primary requirement) other requirements could be a basic idea of how gatsby works.
> 
> I wonder if you have your setup done. ? 🤔 

No sir, I didn't knew about it. Sorry!

[...]

PR #3392, opened 2022-10-20
correct the position of image issue#3338

aditi5050:
the solution to the Image is not aligned with the text on the page: /projects/nighthawk #3338

@legendarykamal please have look

### DIFF
--- a/src/sections/Projects/Nighthawk/index.js
+++ b/src/sections/Projects/Nighthawk/index.js
@@ -93,6 +93,9 @@ const Projects = ({ theme }) => {
         <div className="explain-2">
           <Row>
             <Col lg={6} md={6} sm={12}>
+              <br>
+              <br>
+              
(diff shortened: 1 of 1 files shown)


legendarykamal:
@Nikhil-Ladha or can you please approve the workflow.

Nikhil-Ladha (changes requested):
NO!
This is not how you fix issues like these. Please use CSS

Nikhil-Ladha:
Also, when opening a PR you see a template on how the description should be. Obey it and have the description updated like that.
It is there for a reason.

legendarykamal:
yes , I too think that this will not fix the issue. I doubted so I asked for workflow thanks @Nikhil-Ladha for stepping in .

aditi5050:
yeah sure I'll work on the CSS part and make a PR according to the required description

PR #3507, opened 2022-12-06
Performance: Fix flash of banner on initial load

Nikhil-Ladha:

**Description**
Fix the flash of the banner during the initial load.
Also, due to this flash (which happened because of re-rendering), the meshmap video in the banner was getting downloaded twice. Fixed it with this PR.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -1,0 +1,49 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function HTML(props) {
+  return (
+    <html {...props.htmlAttributes}>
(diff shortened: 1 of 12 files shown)


Nikhil-Ladha:
This may seem like a hack, but with SSR this is the only way through which we can avoid such flashes.

l5io:
🚀 Preview for commit 9613fd442ae1f9d6019e63eaf1c179eb2db16233 at: https://638f430fd6478229bc2f4876--layer5.netlify.app

PR #4211, opened 2023-05-10
Fix for occasional flash of white in the background upon page load

randychilau:
**Description**

This PR fixes a white flash that can happen upon page load in components that do not have a background-color specified.

The examples below uses the projects page in dark mode because light mode and the banner on the home page makes it difficult to see the flash when it occurs:

Multiple [hard refreshes](https://fabricdigital.co.nz/blog/how-to-hard-refresh-your-browser-and-clear-cache) (the flash does not happen all the time):
intermittent-flash

Unfortunately while I found a fix/workaround, I cannot explain exactly why it works (hopefully someone can help on this) .

----

Fix:

- there is a wrapping gatsby element generated (#___gatsby) that can be targeted for styling:
image

- update the globalStyle, and target this element:
image

flash-fix2
(note: any white flash is the browser taking time to reload the page indicated by the blank page during the flash)

----

**Notes for Reviewers**

Here are my testing notes (please let me know other ways you would have tested this):

- I could only reproduce in production builds.

- if the body's background-color does not use props or variables, but a value like "red", it will not show any flash
absolute_value

- the flash is not of the default theme ("light mode"). I changed the light mode and dark mode to have background color of `red` and `blue`, and multiple hard refreshes still show a flash of white:
redblue_white_flash

- in the gifs above, you will notice that all the content appears and the navigation header always loads with the correct background-color, while the other components show the flash. What is different about the header is that it specifies the background-color in its style sheet:

`src/sections/General/Navigation/navigation.style.js`
image

and in this example, the project page, does not specify a background-color:
`src/sections/Projects/Project-grid/projectGrid.style.js`
image

also note that we specify the background-color of the `body` element in the globalStyle:
`src/sections/app.style.js`
image

>**What happens when an element does not specify a background-color?**
>An element that does not specify a [background is transparent (its initial value)](https://developer.mozilla.org/en-US/docs/Web/CSS/background-color#formal_definition). Background-color is also non-inheritable.
image

>This is to allow the background color of the body or parent element to show through the transparent element. And if there is no background color set on the body, then the [browser's background will show](https://dev.to/bcalou/why-you-should-always-set-a-background-color-2gb1).

Since the header appears correctly with each page load and it specifies the background-color using a variable, this shows that there is no issue or delay in the correct values resolving in the styled-components.

----

Best guess:

- the white flash that is appearing is the browser's default background (white).
- the browser's background is showing because the `body` tag is correctly being assigned the styling from the globalStyle, but when it is a variable instead of an absolute value, there **can** be a delay in resolving the variable for some reason.

If anyone has any ideas, explanations, or insights on this issue, please let me know as I felt like I was chasing a ghost in the machine.

----

### DIFF
--- a/src/sections/app.style.js
+++ b/src/sections/app.style.js
@@ -21,6 +21,11 @@ const GlobalStyle = createGlobalStyle`
     outline: none;
   }
 
+  #___gatsby {
+    background: ${(props) => props.theme.body};
+    transition: 0.8s cubic-bezier(0.2, 0.8, 0.2, 1);
+  }
+
 html{
     box-sizing: border-box;
     -ms-overflow-style: scrollbar;


l5io:
🚀 Preview for commit 67af0934a515f9b28065fd0988cbd2e1ca48ab01 at: https://645c0e6f4357f2057c0bc80b--layer5.netlify.app

leecalcote:
⚠️ if you have epilepsy, do not review this PR.

leecalcote:
> ⚠️ if you have epilepsy, do not review this PR.

Bad joke. I know.

randychilau:
well, it is a valid warning.

image

sidenote: I am going to stop using animated gifs and use webm videos to give the viewer some control instead of the PR looking like a bunch of popup ads.

ayushthe1:
@randychilau   Let's discuss this on websites call. Add this as an agenda item in meeting minutes if you would :).
I'll add this item for you if it's ok.

PR #4497, opened 2023-07-06
replace-above-with-below-in-DCO-check-above

vishalvivekm:
**Description**

- It should be "To amend the commits in this PR with your signoff using the instructions provided in the `DCO check below`."
not `DCO check above`
Screenshot 2023-07-07 030513

**Notes for Reviewers**

### DIFF
--- a/.github/label-commenter-config.yml
+++ b/.github/label-commenter-config.yml
@@ -25,11 +25,11 @@ labels:
     labeled:
       issue:
         body: "🚨 Alert! Git Police! We couldn’t help but notice that one or more of your commits is missing a sign-off. _A what?_ A commit sign-off (your email address).\n\n
-         To amend the commits in this PR with your signoff using the instructions provided in the DCO check above. \n\n
+         To amend the commits in this PR with your signoff using the instructions provided in the DCO check below. \n\n
          To configure your dev environment to automatically signoff on your commits in the future, see [these instructions](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)."
         action: open
       pr:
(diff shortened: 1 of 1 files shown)


l5io:
🚀 Preview for commit 2cb01c99a59f5c8a9cbbdcd2f341ca71e0bc0c0a at: https://64a7395872e3300809ad13f4--layer5.netlify.app

Nikhil-Ladha (approved):

leecalcote:
Not quite. Think more about this and you'll realize either above or below is possible.

If we're going to bother with us, I suggest changing the verbiage to remove that either of those prepositions. And then, propagating this change across the other repos.

Nikhil-Ladha:
> Not quite. Think more about this and you'll realize either above or below is possible.
> 
> If we're going to bother with us, I suggest changing the verbiage to remove that either of those prepositions. And then, propagating this change across the other repos.

In case, of an issue it can be `above` but for a PR it has to be `below`.
Anyway, removing the verbiage makes more sense as we are anyway providing a link to set the DCO just below the statement.
@vishalvivekm please make the suggested update (across all the repos, if you would or open issues for the same)

vishalvivekm:
> > Not quite. Think more about this and you'll realize either above or below is possible.
> > If we're going to bother with us, I suggest changing the verbiage to remove that either of those prepositions. And then, propagating this change across the other repos.
> 
> In case, of an issue it can be `above` but for a PR it has to be `below`. Anyway, removing the verbiage makes more sense as we are anyway providing a link to set the DCO just below the statement. @vishalvivekm please make the suggested update (across all the repos, if you would or open issues for the same)

Thanks @Nikhil-Ladha @leecalcote , I'll remove the prepositions ( above/below) in all the repos.

Issue #4498, opened 2023-07-06
Prevent signup using anonymous mail domains


leecalcote:
#### Current Behavior
Anyone can signup to use MeshMap. Signup is necessary to ensure accountability given the open cluster nature of the Meshery Playground.

- https://layer5.io/cloud-native-management/meshmap
- https://layer5.io/cloud-native-management/meshmap/design
- https://layer5.io/cloud-native-management/meshmap/visualize
- https://layer5.io/cloud-native-management/meshmap/collaborate

#### Desired Behavior
Disallow signups from the following domains:

- @duck.com
- @anonaddy.me
- @protonmail.com
- @tuta.io

#### Implementation
Field validation using javascript and regex.

#### Acceptance Tests
1. Users who signup with any of the domains listed above are not allowed to submit the form.
1. Users who signup with any of the domains listed above are shown the following message:

"Use of the <matching-domain-name-here> is not allowed. Please use a non-anonymous email address / domain. 

The Meshery Playground is connected to live Kubernetes cluster(s) and allows users full-control over those clusters. Requiring users to register a non-anonymous user account keeps the Playground safe and healthy for all to enjoy. Without requiring users to sign-in, the Meshery Playground would allow anonymous access for anyone and everyone to potentially perform nefarious acts. Sign-up to receive a user account, entitled with perpetually free and full access the Playground environment."

---
<img src="https://raw.githubusercontent.com/layer5io/layer5/master/.github/assets/images/layer5/5-light-small.svg" width="24px" align="left" /><h2>Contributor Resources and <a href="https://layer5.io/community/handbook">Handbook</a></h2>

The layer5.io website uses Gatsby, React, and GitHub Pages. Site content is found under the [`master` branch](https://github.com/layer5io/layer5/tree/master).
- 📚 See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md).
- 🎨 Wireframes and designs for Layer5 site in [Figma](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs).
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](https://slack.layer5.io).

<img src="https://raw.githubusercontent.com/layer5io/layer5/master/.github/assets/images/buttons/community.webp" height="22px" align="left" />Join the Layer5 Community by submitting your [community member form](https://layer5.io/newcomer).

[...]

PR #4531, opened 2023-07-11
Added email validations on Meshmap Signup page.

saaagarsingh:
I've added necessary email validations on Meshmap signup page. It validates for invalid emails and a given set of anonymous emails addresses/domains.

This PR fixes #
fixes : #4498 
**Notes for Reviewers**
Before:
There was no validation at all, even an invalid email address was getting accepted.

After:

emailvalidation2
emailvalidation1

### DIFF
--- a/src/components/CommonForm/index.js
+++ b/src/components/CommonForm/index.js
@@ -23,18 +23,14 @@ const CommonForm = ({ form, title, submit_title, submit_body }) => {
   const [org, setOrg] = useState("");
   const [occupation, setOccupation] = useState("");
   const [role, setRole] = useState("");
+  const [errorEmail, setErrorEmail] = useState("");
   // const [google, setGoogleAccount] = useState("");
   // const [github, setGithubAccount] = useState("");
   // const [twitter, setTwitterAccount] = useState("");
   // const [linkedin, setLinkedinAccount] = useState("");
-
-
   const errorRole = "Please select role as applicable";
-  const errorEmail = "Please enter a valid email address";
-
   const confirmationMessageRef = useRef(null);  //set reference to confirmation message
   const navBarOffset = 120;
-
   const scrollElementIntoView = (element, offset) => {   //function to bring the confirmation message into view after submittion of form
     var elementPosition = element.getBoundingClientRect().top;
     var offsetPosition = elementPosition + window.pageYOffset - offset;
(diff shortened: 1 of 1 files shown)


l5io:
🚀 Preview for commit f54b31cb1334eba0a940b29b3e453273b5dd28d4 at: https://64ad88bb7f207e1838e826ac--layer5.netlify.app

src/components/CommonForm/index.js
```diff
@@ -44,14 +40,44 @@ const CommonForm = ({ form, title, submit_title, submit_body }) => {
...
+    }
+  };
   useEffect(() => {
     if (submit) {
       axios.post("https://hook.us1.make.com/s4jo47a8ydq6uk7gg16wuukgr4l52p8c", {
         memberFormOne,
       });
     }
-  }, [submit]);
+  }, [submit, errorEmail, validateEmail, validateRole]);
```
codingmickey, line 80: I couldn't get why these additional states needed to have an effect. Like whenever I submit a form only then the inner if will run and the request will be made..

src/components/CommonForm/index.js
```diff
@@ -44,14 +40,44 @@ const CommonForm = ({ form, title, submit_title, submit_body }) => {
...
+      return true;
+    }
+  };
+  /*
+  Validating Email ----
+  1. Check for email if it satisfies the standard forn for email
+  2. Check for anonymous email address/domains. Currently It is checking for only 4 domains, we can add more to regex expression as we move ahead.
+  */
+  const isValidEmail = (value) => {
+    const validEmail = /^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$/.test(value);
```
codingmickey, line 59: Also, the email value is already checked whether it is valid is not via the HTML tag type="email" so adding this additional regex to check the email wouldn't be necessary I felt.

codingmickey (commented):
Hi @saaagarsingh I've reviewed the PR it covers the overall functionality required! Just some minor things that I wanted to discuss

And wanted to point out one thing that @leecalcote mentioned in the issue description about the message.. I felt the whole paragraph had to be included in the error message notifying the users about the policy for why such an email is prohibited.

Some things I would have chosen to do differently..
- Make a common error message tag/state to display the message as this would be helpful in the future for adding different validations (not required for now)

src/components/CommonForm/index.js
saaagarsingh: I added this since useEffect does re-rendering of the component so if the validation fails then submit's value is not gonna change and the component won't get re-rendered.

src/components/CommonForm/index.js
saaagarsingh: You are totally right it wasn't needed but we can say it just adds another layer of assurance that people will correct email address. Also if we go by the regex given in pattern attribute sagar@singh.com5 it will be an invalid email address since there is 5, so the additional check will detect this as well.

src/components/CommonForm/index.js
codingmickey: You're not getting my point

Let me ask you some simple questions..
1. When do you want this post request to be made?
2. When is the value of the `errorEmail` being set?

codingmickey: > I added this since useEffect does re-rendering of the component so if the validation fails then submit's value is not gonna change and the component won't get re-rendered.

Also, the error message is in a state so whenever it changes there is going to be a re-render.. it doesn't need to be given an effect.

vishalvivekm:
@saaagarsingh 

Let's discuss it on the websites call.
Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit#heading=h.lohhtewfwima), if you would :)

leecalcote:
@Shivam-AfA do you have feedback here?

leecalcote:
@saaagarsingh @codingmickey where did we land on this?

saaagarsingh:
@leecalcote As per the said requirements email and roles validation is implemented. Do we need to add validation to any other field??

leecalcote:
@codingmickey, your review here would be helpful...

leecalcote:
@saaagarsingh is this using the same validation logic as the play.meshery.io site is? If not, will you highlight its differences? This will aid in speeding up review. Thanks much for this.

codingmickey:
> @codingmickey, your review here would be helpful...

Sorry for replying late.. but just one thing I wanted to put emphasis on was ...

> And wanted to point out one thing that @leecalcote mentioned in the issue description about the message.. I felt the whole paragraph had to be included in the error message notifying the users about the policy for why such an email is prohibited.

Apart from that this looks good! 🚀

leecalcote:
> @saaagarsingh is this using the same validation logic as the play.meshery.io site is? If not, will you highlight its differences? This will aid in speeding up review. Thanks much for this.

@saaagarsingh reminder...

Chadha93:
> > @saaagarsingh is this using the same validation logic as the play.meshery.io site is? If not, will you highlight its differences? This will aid in speeding up review. Thanks much for this.
> 
> @saaagarsingh reminder...

@saaagarsingh reminder 2....

leecalcote:
@codingmickey will you consider taking over this PR?

saaagarsingh:
@leecalcote @Chadha93 first of all so sorry for the late reply I've been not doing very well from quite some time recovering slowly. Talking about the PR yeah it uses pretty much similar validation as in meshery repo. And I've not added the whole paragraph description instead two precise lines that tells exactly what's wrong there. I also ran several tests for the email as well and it worked completely fine. Once again a huge sorry to everyone for causing inconvenience.

leecalcote:
Thanks much for checking in, @saaagarsingh. Makes sense.

PR #5167, opened 2023-11-29
Revert "[Fix Screenshots] Artifact Hub under "How it Works See It in Action” section"

leecalcote:
Reverts layer5io/layer5#5166


leecalcote:
I missed the fact that this PR fails to build the site. I'll revert it so that other PRs can land and you can resubmit, @sandramsc

sandramsc:
Thanks @Lee - I pushed an update that should fix the build issues, however since it has been merged I'm not sure if it'll take effect, let me know if it makes sense to create a new PR or if pushing the changes has effectively solved the current build status.

PR #5221, opened 2023-12-20
Embedding component testing

iArchitSharma:
This PR is for testing the Embedding component which currently is not working
NOTE: in static folder a script file should be added which is to be loaded

**Notes for Reviewers**

### DIFF
--- a/src/components/EmbeddedDesignLoader.js
+++ b/src/components/EmbeddedDesignLoader.js
@@ -1,0 +1,28 @@
+import React, { useEffect } from "react";
+
+const EmbeddedDesignLoader = ({ scriptPath, containerId }) => {
+  useEffect(() => {
+    const script = document.createElement("script");
+    script.src = scriptPath;
+    script.type = "module";
+    script.crossOrigin = "anonymous";
+
+    script.onload = () => {
+      console.log("Script loaded successfully");
+    };
+
+    script.onerror = () => {
+      console.error("Error loading script");
+    };
+
+    document.head.appendChild(script);
+
+    return () => {
+      document.head.removeChild(script);
+    };
+  }, [scriptPath]);
+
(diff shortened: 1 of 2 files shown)


l5io:
🚀 Preview for commit a92047312c3a28a687e7351b2c910a13b6f12ee5 at: https://658307a938826a25762699f2--layer5.netlify.app

iArchitSharma:
@aabidsofi19 i don't know why but i am unable to push that file, i have shared the file with you on slack so that you can manually add it

leecalcote:
While we'll have instructions and support for Gatsby specifically, @iArchitSharma please also provide the vanilla React app instructions, too.

iArchitSharma:
For a vanilla React app, keep everything the same, except instead of placing the script file in the static folder, put it in the public folder

leecalcote:
@iArchitSharma what's the update today? We didn't receive one.

iArchitSharma:
@leecalcote, no, I haven't. I thought it needed to function correctly initially.

leecalcote:
@iArchitSharma you can and should deliver iteratively. Produce a working vanilla React version and release. Show progress.

iArchitSharma:
Ok so basically the component that I 've created takes two props: 
`scriptPath` - for providing the path of script file that is to be loaded
`containerId` - its the id of div element which is used by script file to perform its task on that element

Apparently when you use the component the div loads correctly as needed in source:
image

when script is loaded successfully `Script loaded successfully` message is displayed on the console, which is being displayed
but on the browser nothing no diagram is being displayed

- I tried replacing all the script code with a simple console message and it worked the console message is being displayed
image

- I tried formatting the script code but it didn't worked
**What i think is that there is some problem with the script file code but I wonder than how is it working on static sites like HUGO**

- I have also tried using **Blob** object but it had the same result

leecalcote:
Refactor to use new NPM package...?

PR #5643, opened 2024-07-07
Bump React to v18

iArchitSharma:
**Description**

React needs to be update to v18 before updating the Gatsby to v5

**Notes for Reviewers**


sudhanshutech (commented):
after this we need to upgrade all those packages that are compatible with react 18 so that they can be upgraded and maintainable otherwise we will be using lower version packages.

PR #6592, opened 2025-07-01
feat: added `X-Frame-Options` and `CSP` for Prevent Clickjacking

saurabhraghuvanshii:
**Description**
Added `X-Frame-Options` and `CSP` for Prevent Clickjacking also added documentation file explaining the security implementation. 

This PR fixes #6584 

**Notes for Reviewers**
- I have test on local it working.
- npm run build failing there is lots of components in repo is outdated I think it should be updated like ` gatsby-plugin-mdx incompatible` 

**[Signed commits](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #6632, opened 2025-07-14
Revert "feat: added `X-Frame-Options` and `CSP` for Prevent Clickjacking"

leecalcote:
Reverts layer5io/layer5#6592

Didn't take effect.


saurabhraghuvanshii:
Hii @leecalcote can  I know what is problem with this, what should I do

PR #6871, opened 2025-09-23
leecalcote/maintainence/tsparticles

leecalcote:
- **Chore: externalize redirects as separate module from gatsby-node.js**
- **Chore: remove unused tsparticles component and package**


src/utils/redirects.js
```diff
@@ -0,0 +1,10 @@
+// utils/redirects.js
+const fs = require('fs');
+const yaml = require('js-yaml'); // Add to package.json if needed
+
+const loadRedirects = () => {
+  const file = fs.readFileSync('./redirects.yaml', 'utf8');
```
leecalcote:
```suggestion
  const file = fs.readFileSync('src/utils/redirects.yaml', 'utf8');
```

PR #7225, opened 2025-12-05
fix workflows and improve logic

saurabhraghuvanshii:
**Description**
#### Problem
The build configuration logic was inverted, causing confusion about when collections are ignored: 
 - The condition process.env.BUILD_FULL_SITE !== "false" meant that when BUILD_FULL_SITE was undefined, it defaulted to a full build
- make site-full didn't explicitly set BUILD_FULL_SITE=true, relying on the undefined default
- This made the behavior implicit and error-prone 

#### Solution
Updated gatsby-config.js to use explicit logic:
- Changed from `process.env.BUILD_FULL_SITE !== "false"` to `process.env.BUILD_FULL_SITE === "true"`
- Default (undefined) → lightweight build (ignores heavy collections)
- `BUILD_FULL_SITE=false` → lightweight build (ignores heavy collections)

Updated package.json to explicitly set BUILD_FULL_SITE=true in the develop script:
 - Ensures make site-full explicitly triggers a full build
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "url": "https://layer5.io"
   },
   "scripts": {
-    "build": "cross-env NODE_OPTIONS=--max-old-space-size=8192 gatsby build",
+    "build": "cross-env BUILD_FULL_SITE=true NODE_OPTIONS=--max-old-space-size=8192 gatsby build",
     "clean": "gatsby clean && rimraf node_modules",
     "develop": "cross-env BUILD_FULL_SITE=true NODE_OPTIONS=--max-old-space-size=8192 env-cmd -f .env.development gatsby develop",
     "develop:lite": "cross-env BUILD_FULL_SITE=false NODE_OPTIONS=--max-old-space-size=8192 env-cmd -f .env.development gatsby develop",


l5io:
🚀 Preview for commit 93a5fb78f71ee0b6a49ca64a7be87e9daf1b2e6a at: https://69332e57d2d0bb7b2349ed1c--layer5.netlify.app

l5io:
🚀 Preview for commit 06ccacaea5999e28d19187cc60a696824b821851 at: https://6933332e9520bf82aef56523--layer5.netlify.app

gatsby-config.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env node */
 
 const isProduction = process.env.NODE_ENV === "production";
-const isFullSiteBuild = process.env.BUILD_FULL_SITE !== "false";
+const isFullSiteBuild = process.env.BUILD_FULL_SITE === "true";
```
CodexRaunak, line 4: How about if we have all the env variables or at least this env variable in `.env.example `so that it's easy to understand for new contributors.
Else we can have a comment on top of the const  regarding the default behaviour:
`// Default is a lightweight build unless BUILD_FULL_SITE is explicitly true
`
```suggestion
// Default is a lightweight build unless BUILD_FULL_SITE is explicitly true
const isFullSiteBuild = process.env.BUILD_FULL_SITE === "true";
```

leecalcote:
Looks good to me. 👍

leecalcote:
Actually, I fail see why this is needed. The existing logic works as expected in my local builds and it seems to be working fine for the production builds that have been recently published.

Where are you seeing that the `make` targets are not being honored?

saurabhraghuvanshii:
@leecalcote  The issue was in the npm script, not the make target. Before the fix: 

- make build → npm run build → gatsby build (without BUILD_FULL_SITE=true)
- So BUILD_FULL_SITE was undefined → isFullSiteBuild = false → collections were skipped

After the fix: 

- make build → npm run build → gatsby build (with BUILD_FULL_SITE=true)
- So BUILD_FULL_SITE=true → isFullSiteBuild = true → all collections included 

The make target works; the npm script it calls needed to set the environment variable. The fix ensures make build includes all collections in production builds  

I also experienced that — after about 10 minutes of building, the integration starts building suddenly and uses all of my resources.

PR #7342, opened 2026-01-18
fix: add cleanup for scroll event listeners to prevent memory leak

Sanchit2662:
## Summary

This PR fixes a **memory leak caused by uncleaned scroll event listeners** in the Layer5 website.
Both the **Navigation** and **Brand** sections were registering `scroll` listeners on mount without removing them on unmount. Since these components are re-mounted on every client-side navigation in Gatsby, this resulted in **accumulating event listeners**, leading to degraded performance over time.

The fix ensures that scroll listeners are **properly cleaned up**, use a **stable function reference**, and are marked as **passive** for better scroll performance.

---

## Problem Description

Before this change, the components attached a scroll listener like this:

```js
useEffect(() => {
  window.addEventListener("scroll", () =>
    window.scrollY > 50 ? setScroll(true) : setScroll(false)
  );
}, []);
```

### What goes wrong:

* The listener is an **anonymous function**, so it cannot be removed
* No cleanup function is returned from `useEffect`
* In Gatsby, layouts and sections are **unmounted and remounted on every page navigation**
* Each navigation adds a **new scroll listener** to `window`
* Over time, multiple listeners fire on every scroll, each updating state

This causes **unbounded memory growth and increasing CPU usage** during normal browsing sessions.

---

### Before

* Scroll listeners accumulate with navigation
* CPU work grows linearly on every scroll
* Memory usage increases due to retained closures
* Possible React warnings from state updates on unmounted components

### After

* Single scroll listener at all times
* Stable memory usage
* Consistent scroll performance
* Safe long-running browsing sessions

This improves performance for **all users**, especially those browsing docs or community pages extensively.

---

## Checklist

* [x] Yes, I signed my commits

### DIFF
--- a/src/collections/members/yash-sharma-2/index.mdx
+++ b/src/collections/members/yash-sharma-2/index.mdx
@@ -11,7 +11,8 @@ bio: "As a dedicated software developer deeply committed to the world of open-so
 badges:
   - meshery
 meshmate: yes
-status: Active
+emeritus: yes
+status: Inactive
 published: true
 permalink: yash-sharma-2
 ---


Sanchit2662:
Hi @kishore08-07 @leecalcote @Rajesh-Nagarajan-11,
This PR fixes a memory leak caused by uncleaned scroll event listeners accumulating on page navigation
Would appreciate your time to review when convenient. Thank you!

CodexRaunak:
@Sanchit2662 Thank you for your contribution! Let's discuss this during the website call tomorrow at 6:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit?usp=drivesdk), if you would 🙂

l5io:
🚀 Preview for commit a15eb938aba75f477459901c1d937911b302ccde at: https://696e6be45027bd6c2ae1cdb0--layer5.netlify.app

Bhumikagarggg:
@Sanchit2662 Thank you for your contribution! Let's discuss this during the website call today at 6:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit?usp=drivesdk), if you would 🙂

PR #7487, opened 2026-03-11
fix(dev): reduce Gatsby local memory usage

marblom007:
## Summary
This PR reduces local Gatsby development memory pressure by making lightweight development the default path, trimming expensive queries and image work, and bypassing Sharp for immutable assets where Gatsby processing adds little value.

## What changed
- default local development to the `core` lite-build profile and document the new controls
- centralize lite-build collection profile handling in `src/utils/build-collections.js`
- narrow page creation and page query work so excluded collections do not force unnecessary GraphQL or image processing
- add placeholder route/schema handling so excluded collections do not break local development
- disable Gatsby core prefetching in development to avoid route-navigation memory growth
- lower lite-dev Sharp defaults to cheaper placeholders and formats
- allow shared background-image infrastructure to use either Gatsby image data or a plain static `src`
- move selected immutable hero backgrounds to `/static/images/lite`
- replace selected immutable company/project/logo `StaticImage` usage with plain `img` tags

## Before vs current
These are the best available comparison points for this branch.

Before these optimizations:
- local `make site` was reported at `34GB+` RAM
- local dev memory after visiting a small number of pages was reported at `20GB+`

Current state on a fresh `npx gatsby clean` + `npm run develop:lite` session:
- development bundle ready after `255.271s`
- idle combined RSS for the main Gatsby process plus worker children: `8,853,360 KB` (~8.85 GB)
- combined RSS after warming `/`, `/community`, `/learn`, and `/projects`: `8,852,400 KB` (~8.85 GB)
- route warming produced no material memory growth in the latest run

## Validation
- verified `npm run develop:lite` starts successfully and serves `http://localhost:8000/`
- verified representative routes respond after startup
- ran ESLint successfully against the touched code files with an explicit Node binary because the terminal PATH in this session was degraded

## Notes
- the `before` figures come from the previously observed local behavior reported during this investigation, not from a single pre-change instrumented run captured in this PR session
- commit used `--no-verify` only because local hook execution in this shell could not find `sh`; the relevant lint validation was run manually before push

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -23,7 +23,8 @@ const shouldBuildFullSite = isFullSiteBuild();
 const excludedCollections = new Set(
   getExcludedCollections({ isFullSiteBuild: shouldBuildFullSite }),
 );
-const isCollectionEnabled = (collection) => !excludedCollections.has(collection);
+const isCollectionEnabled = (collection) =>
+  !excludedCollections.has(collection);
 
 const { loadRedirects } = require("./src/utils/redirects.js");
 const dev404PageSource =
@@ -259,8 +260,9 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
           fieldValue
         }
       }
-      ${isFullSiteBuild
-    ? `memberPosts: allMdx(
+      ${
+        isFullSiteBuild
+          ? `memberPosts: allMdx(
         filter: {
           fields: { collection: { eq: "members" } }
           frontmatter: { published: { eq: true } }
@@ -310,7 +312,8 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
           }
         }
       }`
-    : ""}
+          : ""
+      }
       singleWorkshop: allMdx(
         filter: { fields: { collection: { eq: "workshops" } } }
       ) {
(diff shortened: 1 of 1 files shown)


hortison:
A positive change for sure. My local `make site` was reported at 34GB+ RAM when I browsed to different pages. 

On the current branch, a fresh `npx gatsby clean` plus `npm run develop:lite` came up in 255.271s, with combined RSS for the main Gatsby process and worker children at 8.85GB idle and basically stayed the same after I visited /, /community, /learn, and /projects, which is effectively flat.

leecalcote:
The query/schema reductions, development prefetch disabling, cheaper lite-dev Sharp defaults, static hero asset bypasses, and the immutable-logo migrations.... all positive.

marblom007:
/ agree 😄

leecalcote:
@marblom007, good collaboration here. Thanks for laying waste to this. Long overdue.

l5io:
🚀 Preview for commit 6856858b8199681f2e3afa7d79332ca60ef7e47d at: https://69b1c54b6fea8d875190f187--layer5.netlify.app

PR #7518, opened 2026-03-18
fix: resolve LFX program page content rendering

rishiraj38:
**Description**

This PR Fixes an issue where /programs/lfx was showing the 2026 title but old 2025/2024 content.

**Before:** 

gatsby-node.js was creating a /programs/lfx page for every single year in a loop, so the content was getting randomly overwritten and stuck on whatever file finished last.

**After**: 
We now use a Map to group the files by slug, explicitly pick the newest year (like 2026), and generate the page only once. We also pass the slug to the React component to keep the dropdown UI and markdown perfectly in sync.

**Notes for Reviewers**


l5io:
🚀 Preview for commit 2d9ce4995bf97481788334471409de532e87ea73 at: https://69ba6d436673b75952eca60f--layer5.netlify.app

YASHMAHAKAL (changes requested):
When using string comparison (>) works for cases like lfx-2024 vs lfx-2023, it breaks when comparing versioned vs un-versioned slugs.

For example, in JavaScript, "/programs/gsod-2024/" > "/programs/gsod/" evaluates to false (because - has a lower char code than /). This causes the older gsod page to incorrectly be set as the "latest" over gsod-2024.

@rishiraj38 see the image attached below, see the link in search-bar and page it's displaying.

---
Image

YASHMAHAKAL:
@rishiraj38 i again verified my suggestion and i think your changes are correct. I didn't account for the fact that the slug strings don't have trailing slashes. Because of that, "/programs/gsod-2024" > "/programs/gsod" actually does evaluate to true in Javascript, so this logic perfectly solves the issue.

rishiraj38:
> @rishiraj38 i again verified my suggestion and i think your changes are correct. I didn't account for the fact that the slug strings don't have trailing slashes. Because of that, "/programs/gsod-2024" > "/programs/gsod" actually does evaluate to true in Javascript, so this logic perfectly solves the issue.

For better and safer results, I believe I should parse it and then compare it to determine the latest slug. What are your thoughts? @YASHMAHAKAL

l5io:
🚀 Preview for commit d4b44a63fa56f9bcede20b414378b58001c6592f at: https://69c02075865a3f54ae4826ba--layer5.netlify.app

saurabhraghuvanshii:
Thanks @rishiraj38 lgtm!!

PR #7711, opened 2026-05-06
fix(resources): render SVG thumbnails in cards

arjunmehta-git:
## Summary
- query `thumbnail_svg` and `darkthumbnail_svg` for the Resources grid
- update the shared card component to prefer raster thumbnails and fall back to SVG assets
- restore missing images for SVG-only resource cards in the Resources listings

## Problem
The Resources page requested only raster thumbnail fields, while a set of published resources define only SVG thumbnail metadata. The shared card component also ignored the SVG fields, so those cards rendered without images.

## Result
Resources cards now render available SVG thumbnails, including dark-mode variants when present.

### DIFF
--- a/src/components/Card/index.js
+++ b/src/components/Card/index.js
@@ -13,22 +13,29 @@ const Card = ({
   fetchpriority = "auto",
 }) => {
   const { isDark } = useStyledDarkMode();
+  const primaryThumbnail = frontmatter.thumbnail || frontmatter.thumbnail_svg;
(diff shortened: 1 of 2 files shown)


leecalcote (commented):
Thanks. Let's see what the deploy preview looks like.

ritzorama (approved):

rishiraj38:
Fixed https://github.com/layer5io/layer5/pull/7808.

PR #7734, opened 2026-05-15
fix: blog list view card layout and image styling

rishiraj38:
**Description**

This PR fixes Blog list view. 

**Notes for Reviewers**

Root cause: The Card component (Card.style.js) has `flex-direction: column `and `min-height: 23rem` hardcoded on .`post-block`. The BlogList wrapper tried to override these from the parent, but it couldn't win because styled-components scopes the Card's styles with a unique class, so the Card's own column layout always takes priority over the parent's overrides.


rishiraj38:
### Before 

- Announcement Page

  Screenshot 2026-05-15 at 9 08 18 PM

- Blog Page in List View

  Screenshot 2026-05-15 at 9 08 55 PM

### After 

- Announcement Page

  Screenshot 2026-05-15 at 9 07 42 PM
  
- Blog Page in List View
  
  Screenshot 2026-05-15 at 9 09 31 PM

src/sections/Blog/Blog-list/blogList.style.js
```diff
@@ -15,24 +15,50 @@ export const BlogPageWrapper = styled.div`
...
+
+		.gatsby-image-wrapper,
+		.old-gatsby-image-wrapper {
+			width: 100%;
+			height: 100%;
+		}
+
+		.gatsby-image-wrapper img,
+		.old-gatsby-image-wrapper img {
+			object-fit: contain !important;
```
saurabhraghuvanshii: @rishiraj38  Please don't use !important, try to find root cause.

saurabhraghuvanshii (dismissed):

src/sections/Blog/Blog-list/blogList.style.js
rishiraj38: The Card component hardcodes flex-direction: column in its own styled-component (CardWrapper). The BlogList wrapper's CSS can't override it because styled-components' scoped class has equal/higher specificity. Fix: pass a listView prop to Card so it conditionally renders flex-direction: row from within its own styles.

Bhumikagarggg:
@rishiraj38 Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://meet.layer5.io/websites-minutes), if you would :slightly_smiling_face:

rishiraj38:
@Rajesh-Nagarajan-11 I think my voice wasn't clear, so here’s the main Crux:

The Card component hardcodes `flex-direction: column` inside its own styled-component (`CardWrapper`). Because of styled-components scoping and specificity, the CSS from the BlogList wrapper cannot reliably override those styles.

The fix is to pass a `listView` prop to the Card component and conditionally apply` flex-direction: row `directly inside CardWrapper itself.

Also, the CSS removed from `blogList.style.js` was mainly added to override the `CardWrapper `styles for list view rendering. Since list view handling now exists directly inside the Card component, those overrides are no longer needed.

Rajesh-Nagarajan-11 (approved):

rishiraj38:
@Rajesh-Nagarajan-11, any more changes required?

Rajesh-Nagarajan-11 (approved):

PR #7738, opened 2026-05-19
Fix Adventures callout width prop

Rajesh-Nagarajan-11:
**Root cause:**

`SectionTitle` reads `$UniWidth`, but the callout passed `UniWidth`, so the width override was ignored and the content collapsed to the default 40%.
The image also appeared outside the card because the button added padding/border and the card didn’t clip overflow.

**Fix:**
- Use` $UniWidth`="100%" so SectionTitle gets the intended width.
- Keep the image inside the card by clipping overflow and removing button padding/border.

**Before:**

image

**After:**

image

### DIFF
--- a/src/sections/Adventures-Callout/discuss.style.js
+++ b/src/sections/Adventures-Callout/discuss.style.js
@@ -8,7 +8,7 @@ const AdventuresWrapper = styled.div`
         font-weight: 500;
     }
     h2 span{
-        color:${props => props.theme.secondaryColor};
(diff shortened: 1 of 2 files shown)


Rajesh-Nagarajan-11 (commented):
Kindly ignore the lint changes — they were automatically modified during the pre-commit process.

rishiraj38 (approved):

Sbragul26 (approved):
LGTM :+1:

saurabhraghuvanshii (approved):

PR #7816, opened 2026-06-22
[CI] Fix invalid reusable workflow call in feature-list.yml

rishiraj38:
## Description

Fixes the failing `feature-list.yml` workflow by removing the invalid `call-build-and-deploy--workflow` job.

**Error:**
Invalid workflow file: .github/workflows/feature-list.yml#L57 workflow is not reusable as it is missing a on.workflow_call trigger

**Root cause:**
 The `call-build-and-deploy--workflow` job was calling `build-and-deploy-site.yml` as a reusable workflow via `uses:`, but `build-and-deploy-site.yml` only has `on: push` and `on: workflow_dispatch` triggers; it is missing the required `on.workflow_call` trigger.

**Fix:**
 Removed the unnecessary job. The `stefanzweifel/git-auto-commit-action` step already pushes a commit to `master`, which automatically triggers `build-and-deploy-site.yml` via its existing `on: push: branches: [master]` trigger. 

### DIFF
--- a/.github/workflows/feature-list.yml
+++ b/.github/workflows/feature-list.yml
@@ -50,9 +50,3 @@ jobs:
           commit_user_email: ci@layer5.io
           commit_author: 'l5io <l5io@users.noreply.github.com>'
 
-  call-build-and-deploy--workflow:
-    needs:
-      - trigger-feature-list
-    name: Build and Deploy Site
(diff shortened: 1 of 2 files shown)


leecalcote (changes requested):
Why are we not building the site any longer so that the fresh data is displayed?

leecalcote (approved):
I see the explanation in the PR description now. 👍

PR #7825, opened 2026-06-25
fix: align INR formatPrice multiplier with declared rate

### LINKED ISSUE #7824: fix: INR currency rate mismatch in formatPrice()
Bug Description

In `src/utils/currencies.js`, the `INR` currency entry declares `rate: 88.32` 
but `formatPrice()` hardcodes `price * 88` as the multiplier.

This causes all INR prices rendered via `formatAndConvertPrice()` to be 
undervalued by ~0.36% compared to the declared rate.

 Inconsistency

- `formatPrice()` uses `price * 88` 
- `formatSliderPrice()` uses `price * rate` (88.32) 
- `EUR.formatPrice()` correctly uses `price * 0.86` matching its `rate: 0.86` 

So the two INR formatting functions return different values for the same input.

Affected File

`src/utils/currencies.js`  line 36

 Fix

Change `price * 88` - `price * 88.32` to match the declared rate, consistent 
with how EUR and the `formatSliderPrice` utility already handle it.

Harishrs2006:
Description:
Fixes #7824

The `INR` currency entry declares `rate: 88.32` but `formatPrice()` hardcoded
`price * 88`, causing all INR prices to display ~0.36% lower than they should.

`formatSliderPrice()` already used `price * rate` correctly. This brings
`formatPrice()` into the same pattern, consistent with how EUR handles it.

Change:
`src/utils/currencies.js` line 36: `price * 88` - `price * 88.32`

### DIFF
--- a/src/utils/currencies.js
+++ b/src/utils/currencies.js
@@ -33,7 +33,7 @@ export const Currencies = {
                 currency: "INR",
                 minimumFractionDigits: roundForDisplay ? 0 : 2,
                 maximumFractionDigits: roundForDisplay ? 0 : 2,
-            }).format(price * 88),
+            }).format(price * 88.32),
     },
 };
 


Bhumikagarggg:
@Harishrs2006 Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the meeting minutes, if you would :slightly_smiling_face:

PR #7865, opened 2026-07-12
fix: resolve Aabid Sofi profile issue and make member status checks case-insensitive

rishiraj38:
### Description
This PR fixes an issue where maintainer Aabid Sofi's profile did not render in the active members directory grid, and addresses rendering style issues on their profile card.

#### Root Causes:
1. Aabid Sofi's profile status in `aabid-sofi/index.mdx` was set to `active` (lowercase `a`), whereas the site logic performed case-sensitive checks for `Active` (capitalized `A`).
2. The case sensitivity also broke the profile card background color and inactive/active border styles.

#### Proposed Changes:
* **aabid-sofi/index.mdx**: Capitalized `status: active` to `status: Active`.
* **MemberList.js**: Updated `status` matching logic to be case-insensitive (`status.toLowerCase() === badge.toLowerCase()`) so that lowercase values do not prevent members from appearing in the grid.
* **ProfileCard.style.js**: Updated styled-components condition checks to match `status` case-insensitively, ensuring consistent profile card style rendering.
* **Member-single/index.js**: Standardized status check to be case-insensitive for rendering the `profile-image-inactive` class.
* **Meshmates/index.js**: Standardized status check to filter active/inactive MeshMates case-insensitively.


banana-three-join (commented):
I think the change to Active is good enough for this PR and instead, a beginner issue can be created to integrate a Helper function which returns the output of the status but with the correct capitalization.

Rajesh-Nagarajan-11 (approved):

PR #7908, opened 2026-07-31
ci: fix failing PR preview workflow by splitting into build and deploy stages

Katotodan:
## Summary

The existing PR Preview Deployment workflow was failing at the checkout step.
This PR replaces it with two workflows that separate untrusted PR code
execution from privileged deployment steps, fixing the failure and closing
a security gap in the process.

- **Build PR Preview** (`pull_request`) — builds the Gatsby site from PR code
  with `contents: read` only, no secrets exposed. Safe for fork PRs.
- **Manage PR Preview Deployment** (`workflow_run`) — deploys the build
  artifact to `gh-pages`, comments on the PR, prunes old previews, and cleans
  up on close. Runs with the write permissions/secrets needed for deployment.

## Why

The old workflow used `pull_request_target`, which checks out and builds PR
code while running with the base repo's permissions and secrets — including
for forks. This is what was breaking at checkout, and it's also an unsafe
pattern: untrusted PR code should never run with write access.

The fix follows the standard secure split: build untrusted code with minimal
permissions in one workflow, then hand off only the *build output* to a
second, trusted workflow that has write access and secrets.

PR number/action are passed via artifact metadata rather than
`github.event.workflow_run.pull_requests`, since that field is empty for
fork PRs. The artifact is downloaded by pattern (not an exact name derived
from that field), so the fork-PR case doesn't silently fail.

## Notes for reviewers

Because `manage-pr-preview.yml` is triggered via `workflow_run`, GitHub only
runs the copy on `master`. This needs to be merged before the full flow can
be tested — a follow-up open/sync/close cycle after merge is the real
end-to-end test, including the fork-PR path.

## Summary by CodeRabbit

* **New Features**
  * Added automated preview builds and deployments for pull requests.
  * Preview links are posted automatically when available, making changes easier to review before release.
  * Preview sites are automatically removed when pull requests are closed.

* **Chores**
  * Improved preview workflow reliability, security, and concurrency management.
  * Added safeguards to ensure previews match the intended pull request and commit.
  * Preview build artifacts are retained temporarily to support deployment and troubleshooting.


.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+    steps:
+    - name: Harden Runner
+      uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+      with:
+        egress-policy: audit
+
+
+    - name: Checkout PR
+      uses: actions/checkout@v6
```
banana-three-join: Use the hashed commit counterpart for v7 instead

.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+      with:
+        persist-credentials: false
+        fetch-depth: 1
+    
+    - name: Setup Node
+      uses: actions/setup-node@v6
```
banana-three-join: Use the hashed commit counterpart for v7 instead

.github/workflows/build-site-preview.yaml
```diff
@@ -0,0 +1,97 @@
...
+      env:
+        GATSBY_PREVIEW: "true"
+        GATSBY_SITE_URL: https://${{ github.repository_owner }}.github.io
+      run: |
+        export PATH_PREFIX=pr-preview/pr-${{ github.event.pull_request.number }}
+        npm run build:preview
+        echo -e "User-agent: *\nDisallow: /" > public/robots.txt
+
+
+    - name: Save Metadata
```
banana-three-join, line 67: The metadata should be saved into its own folder and not into the public folder which is being generated by gatsby. They both handle two different responsabilities and, if these steps get switched around, I believe the Gatsby build will create a new public folder and replace the one which stores the metadata, making this step a no-op.

.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+        GATSBY_SITE_URL: https://${{ github.repository_owner }}.github.io
+      run: |
+        export PATH_PREFIX=pr-preview/pr-${{ github.event.pull_request.number }}
+        npm run build:preview
+        echo -e "User-agent: *\nDisallow: /" > public/robots.txt
+
+
+    - name: Save Metadata
+      run: |
+        mkdir -p public/pr-meta
```
banana-three-join: The head commit sha should also be stored in the metadata so we can use it further up as we keep on iterating on the workflows

.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+        echo "${{ github.event.action }}" > public/pr-meta/action
+        echo "${{ github.event.number }}" > public/pr-meta/number
+
+    - name: Upload build artifact
+      uses: actions/upload-artifact@v4
```
banana-three-join: Same feedback here in regards of the versioning. Use the commit version counterpart for v7

.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+      with:
+        name: pr-preview-${{ github.event.pull_request.number }}
+        path: public/
+        retention-days: 1
```
banana-three-join: The retention-days is too low. Up it to 7 days

.github/workflows/build-site-preview.yaml
```diff
@@ -0,0 +1,97 @@
...
+        echo "${{ github.event.number }}" > public/pr-meta/number
+
+    - name: Upload build artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: pr-preview-${{ github.event.pull_request.number }}
+        path: public/
+        retention-days: 1
+
+  cleanup-metadata:
```
banana-three-join, line 84: Is this job really needed? Artifacts already clean-up after themselves so I'm not sure as to how much value this job brings.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,77 @@
...
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - name: Download Artifact
+        uses: actions/download-artifact@v4
```
banana-three-join: Use the commit version counterpart for v8 instead

.github/workflows/deploy-site-preview.yml
```diff
@@ -0,0 +1,77 @@
...
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - name: Download Artifact
+        uses: actions/download-artifact@v4
+        with:
```
banana-three-join, line 39: The exact name for the artifact which is being downloaded is needed

.github/workflows/deploy-site-preview.yml
```diff
@@ -0,0 +1,77 @@
...
+          pattern: pr-preview-*
+          path: temp-artifact
+          merge-multiple: true
```
banana-three-join, line 42: I believe all of the pr-preview-* files are being evaluated so they can be download and put in the temp-artifact path but with `run-id` limits it to the correct artifact from the pr's build run. 

Use the exact name of the artifact instead since the `run-id` only the exact uploaded artifact gets downloaded.

.github/workflows/deploy-site-preview.yml
```diff
@@ -0,0 +1,77 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+          
+            
+      # Extract PR action from metadata uploaded by the build workflow
+      - name: Read Metadata
```
banana-three-join, line 47: Verify the integrity of the metadata such as:

```
number=$(head -n1 ./docs-preview-build/pr/number | tr -d '[:space:]')
          case "$number" in ''|*[!0-9]*) echo "::error::invalid PR number"; exit 1 ;; esac

          action=$(head -n1 ./docs-preview-build/pr/action | tr -d '[:space:]')
          case "$action" in
            opened|reopened|synchronize|closed) ;;
            *) echo "::error::invalid action: $action"; exit 1 ;;
          esac

          echo "PULL_REQUEST=$number" >> "$GITHUB_OUTPUT"
          echo "ACTION=$action" >> "$GITHUB_OUTPUT"
```

.github/workflows/deploy-site-preview.yml
```diff
@@ -0,0 +1,77 @@
...
+          egress-policy: audit
+
+      - name: Download Artifact
+        uses: actions/download-artifact@v4
+        with:
+          pattern: pr-preview-*
+          path: temp-artifact
+          merge-multiple: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
```
banana-three-join, line 44: I believe `actions: read` permission is needed above for the `run-id`

.github/workflows/build-priview-site.yaml
```diff
@@ -0,0 +1,97 @@
...
+        fetch-depth: 1
+    
+    - name: Setup Node
+      uses: actions/setup-node@v6
+      with:
+        node-version: "22"
+        cache: 'npm'
+    
+    - name: Gatsby Cache
+      uses: actions/cache@v4
```
banana-three-join: bump the version to the commit counterpart for v6

banana-three-join (changes requested):

.github/workflows/build-site-preview.yaml
Katotodan: Good question — it's not about cleaning up the artifact itself (agreed, 
retention-days handles that). It exists because the `build` job explicitly 
skips on close (`if: action != 'closed'`), but "Manage PR Preview Deployment" 
still triggers via workflow_run regardless. Without this job, there'd be no 
artifact for it to download on close, so it'd never know to run the removal 
branch.

.github/workflows/deploy-site-preview.yml
Katotodan: Good catch on run-id already scoping this, you're right that pattern was 
overkill for narrowing scope. The reason I avoided an exact name originally 
is that building it requires the PR number up front (e.g. from 
github.event.workflow_run.pull_requests[0].number), which is empty for PRs 
from forks.
As the pattern will substitute the name and there is only one artifact inside, so I think that will work.
Correct me if I'm wrong. :)

banana-three-join (approved):

banana-three-join (approved):

PR #7941, opened 2026-08-08
[Button] Fix Next/Previous buttons staying in clicked state

### LINKED ISSUE #7940: [Bug] Next and previous button stays in clicked state
Description

After clicking the Next button, its color changes from yellow to dark and stays that way. It only returns to the original yellow color after clicking somewhere else on the page, making it look inactive or disabled.

Expected Behavior

After the Next button is clicked, it should return to its normal yellow appearance immediately without requiring the user to click elsewhere on the page.

### Screenshots

https://github.com/user-attachments/assets/eace8c17-4634-4721-a28b-c4787ba3f4d7

### Environment:
- Host OS:
- Browser:

---
<h2>Contributor Resources and <a href="https://layer5.io/community/handbook">Handbook</a></h2>

- 📚 See [contributing instructions](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md).
- 🎨 Wireframes and [designs for Layer5 site](https://www.figma.com/file/5ZwEkSJwUPitURD59YHMEN/Layer5-Designs) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.meshery.io) and [Community Slack](https://slack.layer5.io).

Join the Layer5 Community by submitting your [community member form](https://layer5.io/newcomer).

vermarjun:
**Description**

This PR fixes #7940

**Notes for Reviewers**

The stuck state comes from `reusecore/Button`, not from the pagination component.

The base `ButtonStyle` rule paints `:hover` and `:focus` with the same declaration:

```js
&:hover,
&:focus {
    color: white;
    background: ${props => props.theme.activeColor};   // #3C494F in both themes
}
```

The three variant blocks below it all re-declare `:hover`. None of them re-declare `:focus`. A mouse click leaves the button focused, so the base `:focus` rule keeps winning and the button sits at `#3C494F` until focus moves. Clicking anywhere else blurs it, which is the "only comes back after clicking somewhere else" part of the report.

The change scopes that base rule to `:focus-visible`. Browsers don't match `:focus-visible` when a button is clicked with a pointer, so the button drops straight back to its own color. They do match it on tab, so keyboard users keep the same focus indicator they had before.

Measured on `/resources` with `gatsby develop`, reading computed `background-color` after a real click with the pointer moved off the button:

| button | at rest | after click, before | after click, with fix |
| --- | --- | --- | --- |
| Next `$primary` | `#EBC017` | `#3C494F` | `#EBC017` |
| Previous `$secondary` | `#00B39F` | `#3C494F` | `#00B39F` |

Tab focus still resolves to `#3C494F` on both, so nothing is lost for keyboard navigation.

Before, after clicking Next:

After:

Two notes:

1. Next and Previous were only the visible case. The rule lives on the shared button, so every `$primary`, `$secondary` and `$outlined` button on the site held the dark state after a click.
2. `btn.style.js` isn't prettier formatted on `master`, so the `lint-staged` pre-commit hook rewrites the whole file (75 insertions, 71 deletions) alongside the fix. I committed with `--no-verify` to keep this reviewable at one line. Happy to push the reformat as a separate commit if you want it.

## Summary by CodeRabbit

* **Style**
  * Updated button focus styling to appear during keyboard navigation, reducing focus-ring display during mouse or touch interactions.

### DIFF
--- a/src/reusecore/Button/btn.style.js
+++ b/src/reusecore/Button/btn.style.js
@@ -1,85 +1,89 @@
 import styled, { css } from "styled-components";
-const ButtonStyle = styled.button` 
-    cursor: pointer;
-    display: inline-flex;
-    align-items: center;
-    justify-content: center; 
-    font-family: inherit; 
-    font-size: 16px;
-    text-decoration: none;
-    text-transform: capitalize; 
-    border: 0; 
-    min-width: 170px;
-    padding: 14px;
-    border-radius: 5px;
-    -webkit-transition: 450ms all;
-    transition: 450ms all;
-    position: relative;
-    color: ${props => props.theme.white };
-    background-color: #00B39F;
-    z-index: 999;
-    &:hover,
-    &:focus-visible {
-        color: white;
-        background: ${props => props.theme.activeColor}; 
-        box-shadow: 0 2px 10px ${props => props.theme.whiteFourToBlackFour};
-    }
-    &:active{
-        box-shadow: 0 2px 10px ${props => props.theme.blackFourToWhiteFour};
+const ButtonStyle = styled.button`
+  cursor: pointer;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  font-family: inherit;
+  font-size: 16px;
+  text-decoration: none;
+  text-transform: capitalize;
+  border: 0;
+  min-width: 170px;
+  padding: 14px;
+  border-radius: 5px;
+  -webkit-transition: 450ms all;
+  transition: 450ms all;
+  position: relative;
+  color: ${(props) => props.theme.white};
+  background-color: #00b39f;
+  z-index: 999;
+  &:hover,
+  &:focus-visible {
+    color: white;
+    background: ${(props) => props.theme.activeColor};
+    box-shadow: 0 2px 10px ${(props) => props.theme.whiteFourToBlackFour};
+  }
+  &:active {
+    box-shadow: 0 2px 10px ${(props) => props.theme.blackFourToWhiteFour};
+    transform: scale(0.98);
+  }
+  .icon-left {
+    margin-right: 8px;
+  }
+  .icon-right {
+    margin-left: 8px;
+  }
+
+  ${(props) =>
+  props.$outlined &&
+    css`
+      background: transparent;
+      border: 2px solid ${(props) => props.theme.whiteToBlack};
+      color: ${(props) => props.theme.whiteToBlack};
+      &:hover {
+        box-shadow: 0 2px 10px ${props.theme.whiteFourToBlackFour};
+      }
+      &:active {
+        background: ${(props) => props.theme.lightGrey};
+        box-shadow: 0 2px 10px ${props.theme.blackFourToWhiteFour};
         transform: scale(0.98);
-    }
-    .icon-left{
-        margin-right: 8px;
-    }
-    .icon-right{
-        margin-left: 8px;
-    }
-
-    ${props => props.$outlined && css`
-        background: transparent;
-        border: 2px solid ${props => props.theme.whiteToBlack};
-        color: ${props => props.theme.whiteToBlack};
-        &:hover{
-            box-shadow: 0 2px 10px ${props.theme.whiteFourToBlackFour};
-        }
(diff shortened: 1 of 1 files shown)


vermarjun:
@leecalcote could you take a look when you get a chance? One line in `reusecore/Button`, swapping `:focus` for `:focus-visible` so a mouse click stops parking buttons in the dark `activeColor` state.

Two calls I'd like your read on:

- It lands on every `$primary`, `$secondary` and `$outlined` button, not just the pagination pair in #7940, because the rule sits on the shared component.
- I skipped the prettier hook so the diff stays at one line. Running it reformats all 86 lines of `btn.style.js`, which isn't prettier formatted on `master` today. Say the word and I'll push that as its own commit.

leecalcote:
> @leecalcote could you take a look when you get a chance? One line in `reusecore/Button`, swapping `:focus` for `:focus-visible` so a mouse click stops parking buttons in the dark `activeColor` state.
> 
> Two calls I'd like your read on:
> 
> * It lands on every `$primary`, `$secondary` and `$outlined` button, not just the pagination pair in [[Bug] Next and previous button stays in clicked state #7940](https://github.com/layer5io/layer5/issues/7940), because the rule sits on the shared component.
> * I skipped the prettier hook so the diff stays at one line. Running it reformats all 86 lines of `btn.style.js`, which isn't prettier formatted on `master` today. Say the word and I'll push that as its own commit.

Sure, @vermarjun. I'm taking a look...

leecalcote (changes requested):
Thanks!

The $primary, $secondary, $outlined blocks only override :hover, not :focus, yeah?

After this change a keyboard-focused variant button will still show the base activeColor rather than the variant's own color. How does this manifest in other button variants? Do we potentially need to add &:focus-visible to each variant block?

Prettier skipped? If `--no-verify` was used, let's get a follow up commit.

vermarjun:
@leecalcote yes, that's right. `$primary`, `$secondary` and `$outlined` each re-declare `&:hover` and none of them touch `:focus`.

On whether they need their own `&:focus-visible`, I went and checked, and I think the answer is no. Mirroring each variant's hover into focus would make keyboard focus weaker rather than better. Here's what's actually in those blocks today:

| variant | at rest | its own `:hover` |
| --- | --- | --- |
| `$primary` | `#000000` on `#EBC017` | `#000000` on `#EBC017`, box-shadow only |
| `$secondary` | white on `#00b39f` | white on `#00d3a9`, box-shadow |
| `$outlined` | transparent, `whiteToBlack` border | box-shadow only |

`$primary` and `$outlined` hover to the exact colors they already sit at. Hover changes the shadow and nothing else. So if `:focus-visible` inherited that, tabbing to a Next button would leave it yellow with a faint shadow, which fails WCAG 2.4.7 far worse than what's there now.

Two things worth pulling apart here. The dark-on-keyboard-focus behavior isn't something this PR introduces. On master `:focus` already fires on tab, so a tabbed `$primary` goes `#3c494f` today. This change only removes the pointer case. Keyboard focus is identical before and after.

If you'd rather variants kept their own color while focused, I'd write it as an outline instead of a background swap:

```js
&:focus-visible {
    outline: 3px solid ${props => props.theme.secondaryColor};
    outline-offset: 2px;
}
```

Every variant keeps its identity color, and a ring is easier to get right on contrast than recoloring three buttons. But it sets the focus language for the whole site, not just this file. `btn.style.js` is the only `focus-visible` in `src/` right now, against 19 other `&:focus` rules. Happy either way, though my instinct is that belongs in its own PR rather than riding along with a one-line bug fix.

Prettier is done, pushed as cff87617 with the hook running this time.

One thing I ran into while doing it. `npx prettier --write` isn't idempotent on this file. The first pass indents the three variant blocks a level deeper, the second pass pulls them back, and it holds steady after that, so `--check` still fails right after a single write. `.lintstagedrc.js` runs `eslint --fix` then `prettier --write` once each, so the hook can't land on the stable output by itself. Nothing blocking, just worth knowing if this file shows up churny in a future diff.

PR #8052, opened 2026-09-13
perf: cut build memory so the site builds on an 8 GB machine

saurabhraghuvanshii:
`npm run build` needed ~13.4 GB peak (9.8 GB in a single process) and `gatsby develop --full` OOM-crashed outright. Profiling the webpack module graph, the sharp job queue and the per-stage memory timeline turned up four repository-specific causes.

1. The `@mui/icons-material` barrel dominated the module graph. Two pre-bundled dependencies (`@sistent/sistent`, `@sistent/mui-datatables`) import the whole barrel for 17 icons, and webpack resolved and parsed all ~10,800 icon modules in both the browser and the SSR compilation: 10,819 of 18,949 modules. `gatsby-node.js` now aliases the exact specifier to a generated shim that deep-imports only the icons those bundles use, taking the graph to 8,211 modules (-57%). Deep imports are untouched. `npm run generate:mui-icon-shim` regenerates the list by scanning node_modules, so it is derived rather than guessed.

2. Source images were far larger than anything the site renders. 673 images exceeded 1600px (one PNG was 18534x5417) against a maximum rendered width of 1920px, and sharp's cost scales with source megapixels. Capping the long edge at 2000px (1200px for member photos, rendered at 500px) across 362 files cuts total source megapixels 4,603 -> 2,400 (-48%) and 60 MB of repository size, with no change to any rendered size.

3. `gatsby-plugin-purgecss` did nothing. Its own log reported "Removed ~0.00 KB" every build, because `purgeOnly: ["src/"]` matches no CSS the site actually imports. Removed; CSS output is byte-identical.

4. `@mui/material` was in the shared bundle of all 1,348 pages, because `CookieConsent` (rendered from `layout.js`) imported it for a Snackbar. Rebuilt on styled-components, the site's own styling system.

Also: `gatsby-plugin-preload-fonts` was declared but never registered as a plugin, pulling puppeteer and a Chromium download into every `make setup` (-400 MB of node_modules); `scripts/run-gatsby.js` now derives Gatsby's worker count, sharp concurrency and V8 heap ceiling from the RAM the machine actually has (honouring cgroup limits, so it is correct under WSL2 and in containers) instead of hardcoding 8 GB; and 16 unreferenced assets that were byte-identical duplicates of files still in use were deleted.

Measured on the same machine, full production build:

  peak memory   13,445 MB -> 9,959 MB   (-26%)
  peak process   9,845 MB -> 8,319 MB   (-15%)
  build time         535s -> 416s       (-22%)
  webpack modules  18,949 -> 8,211      (-57%)
  app.js          7,784 KB -> 2,259 KB  (-71%)
  total JS         38.8 MB -> 32.3 MB   (-17%)
  .cache            2.1 GB -> 1.7 GB
  pages built        1,348 -> 1,348     (unchanged)

Under a hard 8 GB cgroup limit `npm run build` now completes (peak 8,914 MB) where it previously died with "Ineffective mark-compacts near heap limit". `make site` peaks at 5,405 MB and is serving in 169s under a 6 GB limit, and `make site-full` now reaches a usable dev server instead of OOM-crashing.

**Description**

This PR fixes #

**Notes for Reviewers**


dhruveshmishra (commented):
src/assets/images/careers/Lee-giving-away-free-t-shirts-at-UT-Austin.jpeg
The image added in this files seems to be reversed

Salmaan-M (commented):
(scripts/run-gatsby.js) Could we revisit the 2048 MB minimum heap here? On machines/containers with less than ~3 GB of RAM, usableGb is still floored at 1.5, which means heapMb is forced to at least 2048 MB. This can leave very little memory for Gatsby workers, Sharp/native allocations, and the OS, potentially causing the same OOM/swap pressure this PR is trying to avoid. Would it be safer to derive the minimum heap from the actual available memory budget rather than enforcing a fixed 2 GB floor?

scripts/run-gatsby.js
```diff
@@ -0,0 +1,136 @@
...
+ * a build takes the whole machine down. Every value here is a *ceiling*, and
+ * anything already present in the environment wins, so CI can still opt into
+ * more parallelism.
+ *
+ * Usage: node scripts/run-gatsby.js <develop|build|...> [gatsby args]
+ */
+const os = require("os");
+const path = require("path");
+const { spawn } = require("child_process");
+
```
Salmaan-M: Could we revisit the 2048 MB minimum heap here? On machines/containers with less than ~3 GB of RAM, usableGb is still floored at 1.5, which means heapMb is forced to at least 2048 MB. This can leave very little memory for Gatsby workers, Sharp/native allocations, and the OS, potentially causing the same OOM/swap pressure this PR is trying to avoid. Would it be safer to derive the minimum heap from the actual available memory budget rather than enforcing a fixed 2 GB floor?

dhruveshmishra (commented):
src/assets/images/Community-pictures/D IMG_6521.jpeg
src/assets/images/Community-pictures/F IMG_6936.jpeg
These images are also seem to be tilted

Maanvi212006 (commented):
same as @dhruveshmishra mentioned these images are inverted. Rest all look correct!
 
- src/assets/images/Community-pictures/D IMG_6521.jpeg
Image

- src/assets/images/Community-pictures/F IMG_6936.jpeg
Image

- src/assets/images/careers/Lee-giving-away-free-t-shirts-at-UT-Austin.jpeg
Image

dhruveshmishra (commented):
Many svgs are seems to be posted duplicately ,is it intentional or what 
Like:-
 src/assets/images/learning-path/istio/Prometheus.webp
 src/assets/images/learning-path/istio/jaeger_1.webp

vedant21-ctr (changes requested):
A few things to address:

Image orientation: Lee-giving-away-free-t-shirts-at-UT-Austin.jpeg, IMG_6521.jpeg, and IMG_6936.jpeg are visibly rotated/mirrored compared with the originals. Please investigate why the resize step is altering orientation. EXIF handling is one possible cause worth checking.
Since 362 images were resized, please validate the full set against the originals, rather than only fixing these 3, to ensure there are no other image regressions.
With 262 changed files and CodeRabbit skipping the review due to the file-count limit, consider splitting the image changes from the build/webpack changes. Not a blocker, but it would make both parts easier to review.

arjunmehta-git (approved):
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #8070, opened 2026-09-14
fix(codeblock): pin Copy button, fix spacing, stop nesting pre in pre

leecalcote:
## Symptom

Once fenced code blocks rendered again (#8068), every CodeBlock showed three defects, e.g. on [Claude Code Skills Not Found After npx Install](https://layer5.io/blog/engineering/claude-code-skills-not-found-after-npx-install):

1. An empty dark band above the first line of code.
2. No bottom margin, so the following paragraph sat flush against the block.
3. The Copy button (browser-default grey) drifted 2rem down, over the code, instead of sitting in the corner.

## Root cause

`src/components/CodeBlock/index.js` rendered `<Pre><CopyCode/><Pre className="prism-code">lines</Pre></Pre>`, reusing one styled `pre` (`margin: auto; margin-top: 1em; padding: 0.5em`) for both the container and the code. The inner `pre` inherited `margin-top: 1em` (the band), nothing supplied a bottom margin, and the button was positioned at `top: 2rem`. Nesting `<pre>` in `<pre>` is also invalid HTML.

Consumers had papered over this. The Meshery hero reached in with `pre pre` and `button` tag selectors, and the handbook's `.codes` wrapper used `margin-top: -2rem` to cancel the band.

## Fix

**CodeBlock** (`src/components/CodeBlock/index.js`)
- `CodeBlockWrapper` (a `div`) owns vertical rhythm (`margin: 1em 0`, zeroed as first or last child) and the positioning context. `Pre` has `margin: 0` and even `0.75rem 1rem` padding.
- The Copy button is pinned top-right (inset 0.625rem), centered on the first line (measured within 1px on every consumer), and is never allowed to hide code the reader cannot get back:
  - **Pointer devices**: transparent until the block is hovered or the button is focused, so at rest the full command is visible (this also replaces the old always-on grey button that sat over the code).
  - **Touch devices** (`@media (hover: none)`, no hover to reveal it): the button stays visible and `Pre` keeps a 4.5rem gutter clear of it. That gutter is `margin-right`, not `padding-right`: a scroll container's end padding is not part of its scrollable area in Chrome, so padding still stranded the last characters under the button at full scroll (measured).
- Line numbers are rendered only when a block has more than one line. A lone "1" numbered nothing and cost ~32px of width that one-line shell commands (most handbook and Meshery blocks) now use.
- Colors come from the Night Owl palette and the site theme (`keppelColor` for hover and the `:focus-visible` ring) instead of hardcoded values.
- The monospace font now survives page-level `* { font-family }` rules. The handbook was rendering code in Qanelas Soft.
- Clipboard failures are handled ("Copy failed" plus `console.error` instead of an unhandled rejection), the feedback timer is cleared on unmount, the collapsible and plain variants share one render path, `key` is no longer passed through `getLineProps`/`getTokenProps`, and the button has `type="button"` and `aria-live`.
- `copy-to-clipboard.js` now rejects when the legacy `document.execCommand("copy")` fallback fails instead of resolving as success, and removes its scratch textarea in a `finally`. `CopyValue` (the other caller) was awaiting it with no `catch` at all, so a rejected `writeText` was an unhandled rejection that still reported "Copied"; it now reports "Copy failed" and clears its timer on unmount.
- `CodeBlockWrapper`, `Pre`, `LineNo`, and `CopyButton` are exported so parents can target them as styled-components selectors.

**Consumers**
- **Meshery hero** (`Animated-steps-list/hero/hero.style.js`): targets `${Pre}`, `${LineNo}`, and `${CopyButton}` instead of nested tag selectors, centers the button on the single command, and stacks lines with `flex-direction: column` so a multi-line snippet would not lay out side by side.
- **Handbook** (`Handbook.style.js`): drops the `-2rem` hack, which after this fix pulled code blocks over the list text above them. `.codes` now uses `margin: 1rem 0` and full width under 768px.
- **Debug Envoy Proxy** (2022 post): it imported `Code` but used 11 raw `<pre><code>` blocks, which MDX 2+ does not route through the `pre` mapping, so they rendered unstyled. They now use `<Code>`, shell line continuations are escaped (`\\`, because a bare `\` before a newline in a template literal is silently dropped), `$ ` prompts are removed so Copy yields runnable commands, and the YAML indentation, a `pod/ pod-name` typo, and a malformed `logging? component = debug` URL are fixed.

**Docs**: `src/components/CodeBlock/README.md` now documents props, layout, the MDX template-literal escaping rule, and the override pattern. It replaces a stale external screenshot.

## Watch for

- Other pages that style CodeBlock through `pre` or `button` tag selectors. I found none beyond the two fixed here; new overrides should use the exported components.
- Page-level `* { ... }` rules like the handbook's, which also hit third-party markup.
- Literal `<pre>` in MDX bypasses CodeBlock. Use fenced blocks or `<Code>`.

## Verification

Dev server (`LITE_BUILD_PROFILE=blog` with `BUILD_COLLECTIONS_EXCLUDE=content-learn,service-mesh-books,service-mesh-workshops,workshops`, which includes blog, handbook, Sistent, and Meshery pages). Every consumer was captured before and after with Playwright at 1280px light, 1280px dark, and 390px mobile (plus a hovered capture per page), with measured geometry:

| Metric (desktop / mobile) | Before | After |
| --- | --- | --- |
| Top of block to first line | 32-36px (band) | 12px |
| Last line to bottom of block | 16-18px | 12px |
| Gap to the following paragraph (blog posts) | 0px | 18px / 16px (1em) |
| Copy button top / right inset | 32px / 20px | 10px / 10px |
| Copy button over code at rest | always | pointer: hidden until hover; touch: clear 4.5rem gutter |
| End of a long line reachable | no | yes on every consumer (touch measured at full scroll) |
| Copy button center vs. first line center | -7px | within 1px |

- [x] Copy writes the exact code to the clipboard, including `\` continuations. The label shows "Copied!" and resets.
- [x] Token font is `"Courier New", Courier, monospace` on the handbook and blog pages.
- [x] No new console warnings from CodeBlock. Remaining warnings predate this change: `defaultProps` on SEO, `<h2>` inside `<p>` in the Envoy post, nested `<a>` in the handbook.
- [x] `eslint` clean on changed JS files.
- [x] Re-measured after the review: line numbers hidden on all 1-line blocks, button opacity 0 at rest on pointer devices, and the end of the first line reachable at full scroll on every touch capture (42 captures, 14 consumers x 3 viewports).
- [x] CI build (first revision). Re-running for the revision.

### Review follow-ups

- @leecalcote's "not clear that the resulting changes are an improvement": fair for the first revision. The always-visible opaque button covered the end of one-line commands, and on the handbook it covered more than before because the code correctly switched to a monospace font. Hence the hover-reveal plus touch gutter above, and dropping the pointless "1".
- CodeRabbit, reserve space for the Copy button: done, as the touch gutter (its `padding-right` suggestion does not work in Chrome, see above).
- CodeRabbit, `execCommand` fallback resolving on failure: fixed, in the helper and both callers.
- CodeRabbit, hero flex direction for multi-line code: fixed.

### Before / After

Blog: Claude Code Skills Not Found (fenced block)

Blog: How to disable "How is Claude doing this session?" (#8066, merged during rebase, so after only)

Blog: Debug Envoy Proxy (raw `<pre>` converted to `<Code>`)

Handbook: Contribution

Sistent: Getting Started / Usage (collapsible)

Sistent: About (inline)

Sistent: About (collapsible)

Sistent: Tokens

Sistent: Installation

Sistent: Button component (`sistent-component` template)

Meshery getting started: hero (Animated-steps-list)

Meshery getting started: platforms, Docker

Meshery getting started: platforms, Apple

## Summary by CodeRabbit

* **New Features**
  * Enhanced code blocks with consistent syntax highlighting, line numbers, copy controls, and collapsible sections.
  * Copy controls now provide clear success or failure feedback.
  * Code examples adapt better to smaller screens, including responsive Handbook layouts.

* **Bug Fixes**
  * Corrected formatting and command examples in the Envoy proxy troubleshooting guide.
  * Improved copy handling to report unsuccessful copy attempts.

* **Documentation**
  * Expanded CodeBlock documentation with usage, configuration, layout, and customization guidance.


leecalcote (commented):
It's not clear that the resulting changes here are actually an improvement.

PR #8110, opened 2026-09-21
fix: correct image orientations after optimization

vedant21-ctr:
## Summary

- Corrected the orientation of 3 images affected by the image optimization in #8052.
- Baked the required rotations into the image pixels before resizing.
- Preserved the optimized dimensions and compression.

## Verification

- Verified the images locally.
- Confirmed the expected dimensions and orientation.
- Confirmed EXIF orientation is no longer required.
- Confirmed only the 3 affected images were modified.

Fixes image orientation regressions from #8052.


leecalcote (approved):

leecalcote:
Thanks!

